PR #178, opened 2013-06-12
Add node_exports query type that returns the exports of a given node.js module file

sqs:
This PR adds a new query type, `node_exports`, in the node plugin that returns the exports of a given node.js module file.

This new query type makes it easier for editors to show a symbolic view of node.js modules, and it makes it easier to generate automatic reference graphs and documentation from node.js modules.

The current workaround using the existing public API is to request completions on `module.exports.` in the file context, but (as far as I know) that requires appending a fake code fragment to the file and does not cleanly handle the case where `module.exports` has been reassigned (e.g., to a function).

The query type is described in the updated docs:

> `node_exports` (node plugin only)
> List the exports of a given node.js module.
> 
> Takes a file field to specify the node.js module file whose exports we want to list.
> 
> Returns an object whose `exports` property holds an array of `{name, doc, origin, start, end, type}` objects, which represent the exports of the module. If the module’s `module.exports` was reassigned (e.g., to make the module itself directly callable as a function), then the object in the array corresponding to that reassigned value will have a null `name` property.

Example query and response for the included [test/cases/node_exports/main.js](https://github.com/sqs/tern/blob/6c52bf9b7a1afcb8cf2757da607e969597a49785/test/cases/node_exports/main.js) file:

``` javascript
// query:
{query: {
  type: "node_exports",
  file: "test/cases/node_exports/main.js"
}}

// yields response:
{ exports: 
   [ { name: null, origin: 'main.js', start: 51, end: 66, type: 'f' },
     { name: 'x',
       origin: 'main.js',
       start: 106,
       end: 107,
       type: 'number' },
     { name: 'f2', origin: 'main.js', start: 42, end: 44, type: 'f' },
     { name: 'f1', origin: 'main.js', start: 35, end: 37, type: 'f' },
     { name: 'f3', origin: 'main.js', start: 70, end: 72, type: 'f.f3' } ] }
```

TODOs
1. Respect the `lineCharPositions` option in the response.
2. Accept a similar set of options to `completions`: types, docs, urls, origins, etc.

Questions
1. Are you, @marijnh, interested in merging this PR? Or is it too node-specific?
2. I had to create a new directory, test/cases/node_exports, with my main.js file, since the existing main.js file was used to test other, conflicting situations. Should I instead update runcases.js to run multiple files within a directory (instead of always just main.js)?
3. Is there a good way to accept a similar set of options to `completions` without reimplementing all of that logic?

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -164,8 +164,12 @@
 
       function makeExport(name, val) {
         var type = val.getType();
-        if (type) type = type.name;
-        return {name: name, doc: val.doc, origin: val.origin, start: val.originNode.start, end: val.originNode.end, type: type};
+        var x = {name: name, doc: val.doc, origin: val.origin, type: type && type.name};
+        if (val.originNode) {
+          x.start = val.originNode.start;
+          x.end = val.originNode.end;
+        }
+        return x;
       }
 
       // check for reassignment of module.exports to a func (it could be reassigned to any


marijnh:
Just a heads-up that the `node_exports` query is now simply called `exports`, and works on all module-loading plugins that are based on the new `modules` plugin.

PR #239, opened 2013-09-11
add support for component(1)

sergioramos:
Based on node.js plugin


johntron:
> Some documentation on this would be great! :-)

Yeah, at least a list of options.

PR #247, opened 2013-10-15
Add support local dependencies to Component plugin

OliverJAsh:
Component has [the notion of "local dependencies"](https://github.com/component/component/wiki/Spec#local), which can be defined inside of `component.json`.

I realise that currently [the Component plugin](https://github.com/marijnh/tern/pull/239) does not support these.


marijnh:
I somehow missed this. Pending a rewrite of the components plugin on top of the new module infrastructure, the components plugin has been dropped.

PR #269, opened 2014-01-10
Treat RequireJS dep paths as relative to current file

sqs:
This PR changes the requirejs plugin to treat dependency paths as relative to the current file.

The previous behavior was as follows. Suppose we have `foo.js` and `mydir/bar.js`. If `mydir/bar.js` required `../foo.js`, then the requirejs would try to open `path.join(server.options.projectDir, "../foo.js")`. With this PR, it finds the correct `foo.js`.

Includes a passing test case.

Here's the part of the spec that describes the desired behavior: https://github.com/amdjs/amdjs-api/wiki/AMD#module-id-format-

> Relative identifiers are resolved relative to the identifier of the module in which "require" is written and called.
> ...
> Relative module ID resolution examples:
> * if module "a/b/c" asks for "../d", that resolves to "a/d"
> ...
> The dependencies ids may be relative ids, and should be resolved relative to the module being defined. In other words, relative ids are resolved relative to the module's id, and not the path used to find the module's id.

And an example of a relative RequireJS dep path in the wild: https://github.com/jquery/jquery/blob/6bff3bf7d72383c1458004d4aae454b61caa400f/src/attributes/classes.js#L2

It would be great to get someone who's experienced with RequireJS to take a look at this.

### DIFF
--- a/plugin/requirejs.js
+++ b/plugin/requirejs.js
@@ -81,6 +81,21 @@
 
   var EXPORT_OBJ_WEIGHT = 50;
 
+  var path = {
+    dirname: function(path) {
+      var lastSep = path.lastIndexOf("/");
+      return lastSep == -1 ? "" : path.slice(0, lastSep);
+    },
+    relative: function(from, to) {
+      if (to.indexOf(from) == 0) return to.slice(from.length);
+      else return to;
+    },
+    join: function(a, b) {
+      if (a && b) return a + "/" + b;
+      else return (a || "") + (b || "");
+    },
+  };
+
   infer.registerFunction("requireJS", function(_self, args, argNodes) {
     var server = infer.cx().parent, data = server && server._requireJS;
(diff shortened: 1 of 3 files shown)


marijnh:
I'm not sure how this got dropped without a single comment, but I seem to remember an alternative patch was merged. Can it be closed?

mrjoelkemp:
It's possible to use requirejs proper to resolve partials. I'm not sure if that fits into the context of this plugin. Something like https://github.com/mrjoelkemp/node-module-lookup-amd/blob/51e4f4125f87ef6e0dd97a4a543d5950dda16509/index.js#L49. 

This way, there's no ambiguity about the resolution behavior and it's less code to maintain in tern.

mrjoelkemp:
Just saw how old this PR is :) Closing sounds good.

marijnh:
Agreed!

PR #405, opened 2014-10-09
Add support for Union types, including completion

MetaMemoryT:
NOTE: these modifications should be reviewed in the context of https://github.com/google/tern-closure/pull/34

Some Union type functionality requires tern-closure to be tested, (that's why some of the new tests reside in tern-closure) as the [doctrine](https://www.npmjs.org/package/doctrine) parser is required to parse union types from annotations.


RReverser:
Did this go anywhere in the end?

MetaMemoryT:
I realized I need to get funding to spend anymore time on this.

PR #425, opened 2014-11-25
fullDocs option for doc_comment

### LINKED ISSUE #411: Doc comments plugin
There needs to be an option to turn off the 'feature' of the Doc comments plugin that kills the comments after the first sentence.  When using Tern I always want all of the comments for a function, not just the first sentence.  It would make more sense for an editor's tern plugin to decide if it wants to trim the returned comments after the first sentence.

> Doc comments plugin
> 
> This plugin, which is enabled by default in the bin/tern server,
> parses comments before function declarations, variable declarations,
> and object properties. It will look for JSDoc-style type declarations,
> and try to parse them and add them to the inferred types, and it will
> treat the first sentence of comment text as the docstring for the
> defined variable or property.
> 
> To turn this plugin off, set doc_comment: null in your plugin option.

This is the code that needs to be turned off via an option:

```
 var first = comments[0], dot = first.search(/\.\s/);
 if (dot > 5) first = first.slice(0, dot + 1);
```

Sevin777:
Closes #411 

Updated demo with default value for fullDocs to make it easy for users
to understand. Added documentation to manual.txt (I don't know how to run the Makefile to regenerate the manual.html file - I'm using Windows).

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -97,7 +97,7 @@
     }
 
     var result = comments[comments.length - 1];
-    if (fullDocs) result = result.trim().replace(/\n[ \t]{0,}\* ?/g, "\n");
+    if (fullDocs) result = result.trim().replace(/\n[ \t]*\* ?/g, "\n");
     else{
       var dot = result.search(/\.\s/);
       if (dot > 5) result = result.slice(0, dot + 1);


marijnh:
Just a heads-up: I've removed this option again in fd7a999 and made it the default behavior.

marijnh:
And another heads-up, attached patch makes the default behavior to strip off long text again, and allows you to add a `docFormat: "full"` field to your queries when you want the full docstring (including newlines). This will require a change to your client if you were relying on these full strings.

sevin7676:
Why? Is the full documentation causing performance issues? If not, shouldn't the client decide to trim the string as it sees fit? I'm not attacking your decision, just hoping for a little bit of an explanation as to why...

marijnh:
The client _is_ deciding to trim the string, by not passing that query parameter. The string trimming logic is involved enough (and likely to be refined in the future) that I don't want to duplicate it in every client. Also, it seems a good idea not to change the old default behavior. There's no performance issues, just that people were suddenly seeing @-jsdoc mess in their tooltips, and not happy about it.

PR #434, opened 2014-12-03
Support for go to definition + find type for string module of node requi...

angelozerr:
This PR gives the capability to retrieve the module type for require string module by using typeAt hook : 
- find definition for string module opens the well JS file modules
- find type (for hover info) displays information of the module.

Here a screenshot when string module "body-parser" is hovered (it call "type" query (findTypeAt)) :

ternnodetypeat

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -168,9 +168,31 @@
 
     return {defs: defs,
             passes: {preCondenseReach: preCondenseReach,
-                     postLoadDef: postLoadDef}};
+                     postLoadDef: postLoadDef,
+                     typeAt: typeAt}};
   });
 
+  function typeAt(file, end, expr, type) {
+    if (expr && expr.node.type == 'Literal') {
+      // we are inside string, check if we are inside require() function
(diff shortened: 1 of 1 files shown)


angelozerr:
> Yes, I understand what it does. I just don't think it is very helpful. 

@marijnh I think this feature is helpful : 

 * an other user needs this feature. See https://github.com/marijnh/tern/pull/464 from @bajtos
 * IDE like WebStorm, Netbeans provides those features.

IMHO, I think "go to definition + find type for string" feature is important. I would like to provide those features for tern.java for Requre (see
http://wiki.netbeans.org/NewAndNoteworthyNB801#Require_JS_Support), like Angular  : 

```javascript
angular.module('A');
angular.module('B', ['A // Alt - to Jump To Definition to the module A
```

bajtos:
Yes, the change proposed in this PR is indeed very helpful, as it makes it easy to navigate to the module referenced via `require`. I don't mind which one of #464 and #434 is landed, as long as this feature is implemented.

Issue #445, opened 2014-12-12
Support JSDoc syntax for optional parameters


p-bakker:
Currently only the Google Closure syntax for optional parameters is supported (using = behind the type), but not the JSDoc way (using square brackets around the param name)

dannysu:
This seems to be already resolved, by the issue is still open.

marijnh:
You're right. I guess it got implemented in some other context.

PR #446, opened 2014-12-12
Support "square bracket around @param name inside JSDoc for optional params" notation

p-bakker:
Implements #445

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -280,7 +280,7 @@
         case "type":
           type = parsed; break;
         case "param": case "arg": case "argument":
-            var name = m[2].slice(parsed.end).match(/^\s*(\[?\s*)([^\]\s]+)\s*(\]?).*/);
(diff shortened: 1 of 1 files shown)


sevin7676:
This is not working properly when using optional parameter and default value:

http://usejsdoc.org/tags-param.html
image


## Example of Problem ##
http://ternjs.net/doc/demo.html
image

## Proper Behavior ##
 *My [Tern plugin](http://sevin7676.github.io/Ace.Tern/demo.html) modifies the contents of tooltips by re-parsing the jsdocs. I just upgraded to the latest Tern version which is why I found this issue as I had to update my re-parsing and matching with terns arg hints data due to the `?` now being appended to optional params. I'm NOT suggesting that tern modifies it's response to look like my arg hints, I'm just showing that tern should recognize that the second parameter is optional.*
image

marijnh:
Attached patch should fix the parsing of `[foo=10]` style parameter declarations.

PR #464, opened 2014-12-31
plugin/node: `definition` for `require()`

bajtos:
Enhance the node plugin to decorate string literals passed to `require` calls so that `definition` queries for these literals return results pointing to `module.exports` in the target files.

Example:

```js
var express = require('express');
```

Place the cursor inside the `express` string and invoke `:TernDef` in Vim. With this change in place, Vim navigates to `node_modules/express/index.js`.

This functionality is already available in WebStorm and possibly in other IDEs too.

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -36,6 +36,7 @@
     module.propagate(scope.defProp("module"));
     var exports = new infer.Obj(true, "exports");
     module.origin = exports.origin = origin;
+    module.originNode = exports.originNode = scope.originNode;
     exports.propagate(scope.defProp("exports"));
     var moduleExports = scope.exports = module.defProp("exports");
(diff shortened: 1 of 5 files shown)


angelozerr:
@bajtos it seems that's it's the same PR than https://github.com/marijnh/tern/pull/434

bajtos:
@angelozerr Good catch! I left a comment in the other PR.

marijnh:
Nope, this is not the same pull request as #434. For one, this one leaves the type of the string to be a string, _and_ it includes a test, which is always a good sign. Merged as 90dfc43

angelozerr:
Very cool! It works like a charm.

bajtos:
Nice, thanks for landing my patch :+1:

PR #471, opened 2015-01-05
Move position of doc property

tmcw:
I'm not totally confident that this should be fixed here, or if it's a regression
elsewhere, but I'm seeing the docstring of JSDoc comments living _under_ the
type property.


marijnh:
Can you provide a test case that fails? This fix is definitely not right. Looking at the relevant code, I did spot a suspicious piece of code, which I've fixed in attached patch, and might solve your problem.

PR #475, opened 2015-01-09
ecma6.json 

angelozerr:
This PR provides a first version of ecma6.json.

This def must be improved to support iterator (see TODO_ITERATOR).

I think too that TypedArray_ctor, Map_ctor, Set_ctor should be created and used too.

In other words this PR desribes the whole ES6 defined at https://developer.mozilla.org//en-US/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla

This def must be improved to manage iterator and some !effect.

### DIFF
--- a/defs/ecma6.json
+++ b/defs/ecma6.json
@@ -204,14 +204,6 @@
     "from": "TypedArray.from",
     "of": "TypedArray.of"
   },
-  "Function": {
-    "prototype": {
-      "toMethod": {
-        "!type": "fn()",
-        "!url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toMethod"
-      }
-    }
-  },
   "Int16Array": {
     "!type": "fn(length: number)",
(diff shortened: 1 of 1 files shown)


marijnh:
Looks great. But shouldn't the `Promise` properties live under `Promise.prototype`?

angelozerr:
> Looks great. But shouldn't the Promise properties live under Promise.prototype?

As I said you, it's not perfect and I think it should be improved, but it's a first step. I think we could improve it step by step. What do you think?

More I tell me if tern could customize acorn with 6 version when ecma6.json is loaded.

marijnh:
I think it is safe to simply add support for ES6 syntax by default -- it is almost entirely backwards-compatible anyway. Acorn parses ES6 now, but before we turn that on in Tern we'll have to add support for new syntax constructs.

marijnh:
So should the Promise properties be moved to `prototype`, or is there a reason they are directly on the constructor?

angelozerr:
> So should the Promise properties be moved to prototype, or is there a reason they are directly on the constructor?

To be honnest with you, I don't know -( I have generated that from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

marijnh:
I've merged this as de4efe7df844ea357f08.

Issue #467, opened 2015-01-05
Wrong file position


kimgronqvist:
In emacs (24.4, Windows 8 64-bit) Tern gets gets out of sync in the file after a couple of lines. I get this behaviour in all files, but here's an example from the tern code base.

Here everything is okay:
![image](https://cloud.githubusercontent.com/assets/3421067/5613965/f3887d0e-94eb-11e4-9aa0-37bc92143078.png)

But further down I get this:
![image](https://cloud.githubusercontent.com/assets/3421067/5614138/18ee2344-94ee-11e4-8bcc-f235a54bd88a.png)

I also get message like this one when I try to use tern-find-defintion:
`error in process filter: tern-find-position: Args out of range: #<buffer tern.js>, 35027, 35077`

kiwanami:
Which coding of end of line do you use?
I met similar troubles during editing for `utf8-dos` coding system, too.

It may be caused by the difference of handling EOL between Emacs and node.js, Emacs treats CRLF as one character, but node.js does as 2 characters.
The difference arises when node.js reads source codes which are opened by Emacs or required by some source codes.
If a source code is modified in Emacs, the modified code is sent to node.js from Emacs. In this case, because the EOL codes correspond, I don't meet the problem.

Deleting CR code when file reading, I confirmed that this problem is resolved.
I modified `updateText` function in `lib/tern.js`, like following:
https://github.com/marijnh/tern/blob/9ecb8285580a049b0e91c28cdcc350c02b2f62ee/lib/tern.js#L75
```
  function updateText(file, text, srv) {
    file.text = text.replace(/\r/g,""); // adjust text position to editor
    file.ast = infer.parse(file.text, srv.passes, {directSourceFile: file, allowReturnOutsideFunction: true});
    file.lineOffsets = null;
  }
```

But, I don't know whether this code is OK even in the other text editors, such as vim, ST and browser's editors.

kimgronqvist:
I might have spoken to soon :) Switching to utf8-unix fixes the problem, but the change to updateText doesn't seem to make any difference.

kiwanami:
I may miss some cause...

Will you check the diff? I modified two lines:
```diff
diff --git a/lib/tern.js b/lib/tern.js
index 628a3e0..ddc3886 100644
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -73,8 +73,8 @@
   File.prototype.asLineChar = function(pos) { return asLineChar(this, pos); };
 
[...]

PR #477, opened 2015-01-09
Deleting CR to adjust the text position to editor

kiwanami:
Issue #467
If the source code is encoded in DOS-like line separator,
Emacs reports incorrect type definition or throw errors.
It is caused by the difference of handling EOL between Emacs and
node.js, Emacs treats CRLF as one character, but node.js does as 2
characters.

This patch deletes the CR character from the source codes
which are loaded on the memory and adjust to Emacs handling.

I checked this patch on Emacs 24.3/Linux, Vim 7.4.542/Linux, NTEmacs 24.4/Windows8 and Vim7.4(kaoriya)/Windows8.

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -73,9 +73,9 @@
   File.prototype.asLineChar = function(pos) { return asLineChar(this, pos); };
 
   function updateText(file, text, srv) {
-    file.text = text;
+    file.text = text.replace(/\r/g,""); // adjust text position to editor
     infer.withContext(srv.cx, function() {
-      file.ast = infer.parse(text, srv.passes, {directSourceFile: file, allowReturnOutsideFunction: true});
+      file.ast = infer.parse(file.text, srv.passes, {directSourceFile: file, allowReturnOutsideFunction: true});
     });
     file.lineOffsets = null;
   }


marijnh:
Did you test whether this causes problems for Sublime Text (which does not use row/column positions, but rather single-number character offsets, and thus might get confused when the offsets no longer correspond to raw file offsets).

kiwanami:
I checked the following test for `tern_for_sublime` on ST2/Windows8 and Linux. I have found no problem.

### Test Procedure
1: install normal ternjs plugin (0.7.0)
2-1: open a JS file which is encoded in LF.
2-2: examine tern functions (completion, argument hints and jump definition), and after file editing do again
2-3: open a JS file which is encoded in CRLF.
2-4: examine the same test as 2-2
3: apply this patch to ternjs
4: examine the same test as 2-1~2-4

On Emacs, only test 2-4 fails. However, on vim and ST2 I have found no problem.
I wonder that this patch does not affect the result on vim and ST2.
Should I check another test?

marijnh:
Indeed, I can't find problems with ST either. Merged as 98410a7adfbe, and slightly changed in 9fdb21b801117389 to only delete CRs that are followed by an LF.

angelozerr:
I have tested with Eclipse on Windows and it doesn't work -(

Here my JSON trace : 

 * file posted : 
```json
{"files":[{"name":"test.js","text":"$('foo');\r\n\r\n\r\n","type":"full"}]}
```

 * completion executed at the end of the file : 

```json
{"query":{"type":"completions","file":"test.js","end":13,"types":true,"docs":true,"urls":true,"origins":true,"caseInsensitive":true,"lineCharPositions":false,"expandWordForward":false,"omitObjectPrototype":false,"file":"test.js"}}
```

I have the following error : 

```
Tern error#completions with 13ms: 
tern.TernException: TernError: Position 13 is outside of file.
```

Not that if I remove last `\r\n` it works (no error).

angelozerr:
I think this behaviour depends on the editor, so perhaps it could be a better idea to replace or not text switch a configuration.

angelozerr:
Or perhaps this replace should be done only when lineCharPositions is to set as true?

kiwanami:
@marijnh 
Thank you for merging!
I'm sorry for breaking the eclipse bridge...

@angelozerr 
Yes, I also think that it depends on the editor.
Although switching by configuration is good idea, I think that the users should not touch the config file such as `.tern-project`.
How about the command line argument?

marijnh:
Well, I already tagged 0.8, so this is unfortunate. I'll try to figure out a solution tomorrow, either reverting the patch or doing something to make it configurable, and then bring out another release.

marijnh:
Patch 780ce17 makes this optional, defaulting to off, and has the emacs mode pass the appropriate option when starting the server.

@angelozerr Does this solve your issue?

angelozerr:
> @angelozerr Does this solve your issue?

Yes it works like a charm. That's cool I have nothing todo since `--strip-crs` option is disabled by default. 

Thanks!

kiwanami:
@marijnh @angelozerr 
Thank you!!

PR #478, opened 2015-01-12
Improve document#createElement by returning teh well HTML Element instance

angelozerr:
This PR gives the capability to return the well HTML instance according the createElement parameter : 

 * here a sample with any element : 
```javascript
var elt = document.createElement('any') // here elt is an instance of Element
```

 * here a sample with HTML div : 

```javascript
var div= document.createElement('div') // here elt is an instance of HTMLDivElement
div.a // Ctrl+Space shows align attribute (only available for div).
```

To support that, I needed to create a tern function `Browser_createElement` that I have added inside `def.js`. It's not very clean. I think `brower.json` should be tranformed to a tern plugin `browser.js`.

### DIFF
--- a/defs/browser.json
+++ b/defs/browser.json
@@ -1337,8 +1337,17 @@
     "!url": "https://developer.mozilla.org/en-US/docs/Web/API/HTMLDataListElement"
   },
   "HTMLDivElement": {
-    "!type": "Element",
-    "!url": "https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement"
+    "!type": "fn()",
+    "prototype": {
+      "!proto": "Element.prototype",
+      "align": {
+        "!type": "string",
+        "!url": "https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement",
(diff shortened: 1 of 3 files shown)


marijnh:
I think this is complexity without much gain. The plugin will still be completely ignorant about what kind of element, for example, `document.body.firstChild` should return. Just add the `align` property to `Element`, like I've done with other properties that only occur on some types of elements (such as `getContext` and `value`).

angelozerr:
@marijnh I have created this PR, because I have started to wtrite a plugin which improve browser.json with getElementById

```html
<html>
  <div id="MyId" ></div>
</html>
```

```javascript
var div = document.getElementById('MyId'') 
div. // here Ctrl+Space shows align
```

To do that I load inside a tern plugin HTML with parse5. And once this issue https://github.com/inikulin/parse5/issues/30 will be done I could manage Go definition when user Ctrl+click on "MyId".

marijnh:
I don't see a way to reliably do this -- even if you have a reference HTML document, DOM structure, you won't be able to return appropriate types for things like `firstChild`. Even `getElementById` only works reliably if the script didn't update the DOM. I think it is a better idea to stick to just putting all possible properties on the `Element` type itself, and accepting the dynamic typing as inavoidable here.

PR #479, opened 2015-01-15
invoke completion-at-point with C-TAB

dshamaeli:
First : thank you Marjin for both "Eloquent JavaScript" and Ternjs. I owe you a lot for these.
Second: My tern auto complete didn't work as I expected since the default M-TAB invokes OS application  switch key (I use Ubuntu 14.04). I wanted to use  js3-mode, Auto complete(TAB), Yasnippet  (I put it to shift-TAB so it wont conflict with auto) and Tern (I want to invoke it with C-TAB and . )  so I added 

        (define-key tern-mode-keymap [(control tab)] 'completion-at-point)

in keymap section of code. Which I think will help many who are new to Emacs. 
Now I can get all these great packages together. and who needs WebStorm any more :).

https://github.com/dshamaeli/My-Emacs.git

### DIFF
--- a/emacs/tern.el
+++ b/emacs/tern.el
@@ -537,6 +537,7 @@ list of strings, giving the binary name and arguments.")
       (tern-send-buffer-to-server))))
 
 (defvar tern-mode-keymap (make-sparse-keymap))
+(define-key tern-mode-keymap [(control tab)] 'completion-at-point)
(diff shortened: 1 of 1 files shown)


marijnh:
I think this is something that should go into a user's `.emacs`, not in the Tern plugin (`completion-at-point` is a standard command, and where to bind it is up to the user).

Issue #481, opened 2015-01-15
Proposal: Ability to Extend Existing Query Type


kgritesh:
I am using tern in a project where we would like to provide custom list of completions to the user in some specific situations but use the existing code completions provided by the library otherwise.  My idea is to provide a hook similar to defineQueryType, using which people can register a custom implementation for any of the existing query type. The custom implementation receives the default implementation for that query type as a parameter and hence call the default implementation as well.

I am ready to contribute towards this and can make a pull request, but just wanted to confirm if something like would be acceptable.

marijnh:
So far, we've been taking the approach of providing query-type-specific hooks (such as `completion`) to extend queries, since that allows things like query-specific error checking to happen before the custom code is ran. What is your use case for this feature?

PR #482, opened 2015-01-16
Added a method that allows extending existing query types

kgritesh:
This pull requests add a mechanism to extend existing query types supported by Tern.  It defines a new method which can be used to register a function that is called instead of the default one for any of the existing query types.  This function also receives the default implementation as a parameter and hence the user can build on top of default implementation.  This is in response to the issue https://github.com/marijnh/tern/issues/481

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -66,6 +66,25 @@
 
   exports.defineQueryType = function(name, desc) { queryTypes[name] = desc; };
 
+  exports.extendQueryType = function(name, desc) {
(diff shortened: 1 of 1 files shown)


marijnh:
Apologies for never getting around to commenting on this. I think it is usually better to create a new query type, if you want to do something significantly different. For specific query types that can meaningfully be extended (such as completions and finding a type for something), we've added events that plugins can hook into to do the extension, which is probably a safe/cleaner approach that wholesale wrapping of requests.

PR #494, opened 2015-02-12
Getting code completion to work with custom Types

yGuy:
The gather callback obviously expected the reported obj in the gather callback to be an instance of infer.Obj and directly uses its implementation private "props" property in "obj.props[]". Instead I believe it should use the proper API "obj.getProp()". This bit me when I tried to implement my own static Type that did not have the "props" field. The proposed change makes the implementation accept custom "infer.Type" implementations, too.

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -622,7 +622,7 @@
       completions.push(rec);
 
       if (obj && (query.types || query.docs || query.urls || query.origins)) {
-        var val = obj.props[prop];
+        var val = obj.getProp(prop);
         infer.resetGuessing();
         var type = val.getType();
         rec.guess = infer.didGuess();


marijnh:
There will be a lot of other places that assume the built-in set of types are the only types that exist. This interface is not intended to be extended. What kind of type are you implementing?

yGuy:
I have a large system that describes our types. I found that using the "defs" format was not feasible, since it does not support generics completely. I tried to create types whose properties are queried lazily, because feeding the system with all types at once probably performs awfully.
So I implemented my one type extending infer.Type. 
This works very good until I use code completion. After this fix it seems I can get most functionality to work. However I already found out that using new infer.Obj(myType) will also not work, since it also requires the "props" property. The problem with the props property is that I cannot easily hook into the time when it is queried for the first time. So lazy initialization of my types is almost impossible.
Speaking of which: I *could* make "props" a computed property with a getter function and then initialize it.

Do you have another idea how to implement the system so that types are populated and created lazily?

If you are interested: This is the system I am trying to integrate it with:
http://docs.yworks.com/yfileshtmlv2/index.html

marijnh:
Why don't you include a `props` property (which is, in fact, part of the object's interface)? Keeping most of the `Obj` type and only extending the `getProp` method might work. You'll still have some issues unless you also make sure that the objects' `constructor` property refers to `Obj`, because that is used in several places to make sure code isn't dealing with a function or array.

Eventually, I'd like to add proper support for computed properties (having arbitrary code execute when a property is accessed), but this doesn't exist yet. If you want to sponsor its development, write me.

yGuy:
Regarding "computed properties": I think this was a misunderstanding. I think I understand that tern does not understand getter/setter properties in the code it analyzes. However my idea was to use that Javascript language feature in my own Obj subtype to become notified when the property is accessed and lazily initialize it. I tried that in the meantime and although it's a bit dirty, it seems to work. 
So I am no subtyping Obj just for the purpose of lazily initializing the props property.
Regarding sponsporing: This is indeed something we consider after we have finished our prototype.
Thanks again for a great product!
Note: I still believe that my initial merge request would improve the situation a little, although with my workaround it is not necessary anymore... Feel free to close the request if you don't want to merge it!

PR #506, opened 2015-03-14
Always glob in the root of the project

Slava:
`glob.sync` defaults its search to the current working directory. When used from the Sublime Text 3 plugin, the current working directory becomes the path of the first opened file. This is incorrect, the `loadEagerly` property is intended to load files relative to the root (or `.tern-project` file).

### DIFF
--- a/bin/tern
+++ b/bin/tern
@@ -151,7 +151,7 @@ function startServer(dir, config) {
   });
 
   if (config.loadEagerly) config.loadEagerly.forEach(function(pat) {
-    glob.sync(pat).forEach(function(file) {
+    glob.sync(pat, { cwd: dir }).forEach(function(file) {
       server.addFile(file);
     });
   });


Slava:
Actually this change breaks some other behavior in the ST3 plugin, such as "jump to definition". Please don't mind this.

marijnh:
Does calling `process.chdir` to switch the server's working directory to the project dir work better? I think a bunch of code is written with the assumption that the working dir is the project dir, but as it is, that is indeed not guaranteed (though, if stared by a plugin, I believe it should be the case)

Slava:
Hmm, I am not that familiar with the code to make a judgement call if that change suits better. It could work better, but with only that change everything seems to work fine for me in my ST3.

(The previous comment about ST3 plugin not working was result of my misunderstanding.)

marijnh:
Oh, okay, I assumed you had found a problem with the patch. If not, then sure, this looks good. Merged.

Slava:
Thanks!

PR #509, opened 2015-03-19
Fix an incorrect sourcing of the original file.

Slava:
This looks like a typo, in a block where we acquired start and end positions
based on the spanFile, we return the span.origin as the name of the file.
span.origin and spanFile.name can differ.

I don't understand this part of the code 100%, but I noticed a bug working with my code and running the `definition` command. And then I debugged it down to this statement.


marijnh:
Can you give an example of a situation where they differ? It's been a while since I touched this code, but it seems that they should always be the same.

PR #514, opened 2015-03-26
Add FormData to browser def

othree:
A new global object help user send structured data, both XHR2 and fetch can use this.

https://developer.mozilla.org/en-US/docs/Web/API/FormData


marijnh:
Thanks! Merged as e457341

PR #520, opened 2015-04-07
Omit Object Prototype for completion inside object literal even if query...

angelozerr:
....omitObjectPrototype is set to true

This PR omit Object Prototype for completion inside object literal even if query.omitObjectPrototype is set to true

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -610,7 +610,7 @@
     function gather(prop, obj, depth, addInfo) {
       // 'hasOwnProperty' and such are usually just noise, leave them
       // out when no prefix is provided.
-      if ((ignoreObj || query.omitObjectPrototype !== false) && obj == srv.cx.protos.Object && !word) return;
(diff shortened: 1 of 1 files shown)


lib/tern.js
```diff
@@ -610,7 +610,7 @@
     function gather(prop, obj, depth, addInfo) {
       // 'hasOwnProperty' and such are usually just noise, leave them
       // out when no prefix is provided.
-      if (query.omitObjectPrototype !== false && obj == srv.cx.protos.Object && !word) return;
+      if ((ignoreObj || query.omitObjectPrototype !== false) && obj == srv.cx.protos.Object && !word) return;
       if (query.filter !== false && word &&
```
marijnh: Why `ignoreObj` and not `objLit`? The latter seems to more precisely describe what we are talking about (and it looks like they'll have the same value at this point).

angelozerr: Yes you are right @marijnh . I have updated  my PR with your suggestion.

marijnh:
Great. Merged as 774372e

PR #525, opened 2015-04-20
Mark the type which is coming from the comment

angelozerr:
This PR gives the capability to know if a type is created from a comment or not.

In my case I need this information to support validation with comments https://github.com/angelozerr/tern-lint/issues/30

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -369,6 +369,7 @@
   function propagateWithWeight(type, target) {
     var weight = infer.cx().parent._docComment.weight;
     type.type.propagate(target, weight || (type.madeUp ? WG_MADEUP : undefined));
+     type.type.fromComment = true; // mark the type is coming from the comment
   }
 
   function applyType(type, self, args, ret, node, aval) {


angelozerr:
@marijnh note that this new property `type.fromComment`  could be used too for completion, find type, when type must be found. Like `preferFunction` query option, we could have `preferDoc too query option`.

marijnh:
You really can't just assign to type values like that. They may be shared (it might be the `bool` type, etc).

angelozerr:
@marijnh the type can be created with different mean : 

 * with JSON Type Definition
 * with JSDoc
 * by guess

It should be very cool if we could know which mean has created the type. In my case I need it for validation, but I think we could use too to get the type (with preferDoc, etc).

angelozerr:
@marijnh when you have time, could you give me some direction to know the origin of the type create (from JSON Type Definition, JSDoc comment, guess). Thanks!

angelozerr:
> You really can't just assign to type values like that. They may be shared (it might be the bool type, etc).

You are right, that's why I have created the new PR https://github.com/marijnh/tern/pull/542 which supports for the moment just simple type like boolean, string, number.

PR #532, opened 2015-04-27
String#replace support for string or Regex type.

angelozerr:


marijnh:
Great! Merged as 9f22056

PR #534, opened 2015-04-29
[emacs mode] Fix AC error when type info is null

kiwanami:
When the tern server returns completion items which have null type field, emacs raises following error message:

```
Error running timer `ac-show-menu': (wrong-type-argument stringp nil)
Error running timer `ac-update-greedy': (wrong-type-argument stringp nil)
```

This patch fixes this problem and display the items appropriately.


marijnh:
Thanks! Merged as 7e5f750

PR #544, opened 2015-05-20
Add option to make tern completions synchronous.

renatofdds:
With this option on there is no need to bind a separate key specifically
for tern-ac-complete, it will be queried automatically by auto-complete.


renatofdds:
Actually this bugs when editing arrays, objects or strings. I'll further test it and update it once i find a final solution.

PR #546, opened 2015-05-28
Add option to make tern completions synchronous.

renatofdds:
With this option on there is no need to bind a separate key specifically
for tern-ac-complete, it will be queried automatically by auto-complete
and list completions along with other ac-sources.

### DIFF
--- a/emacs/tern-auto-complete.el
+++ b/emacs/tern-auto-complete.el
@@ -37,6 +37,16 @@
   :type 'boolean
   :group 'auto-complete)
 
+(defcustom tern-ac-sync t
+  "[AC] If t, auto-complete will wait for tern canditates before starting.
+This enables tern canditates to integrate automatically in auto-complete without
+the need for a separate keybinding.
+
+Remember to add ac-source-tern-completion to ac-sources."
+  :type 'boolean
+  :group 'auto-complete)
+
+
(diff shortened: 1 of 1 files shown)


marijnh:
I'm out of my emacs-depth here -- I don't really understand what is being done and why.

@kiwanami Do you have an opinion on this patch?

renatofdds:
Advising functions is the emacs easy way to attach functionality before, around, or after an already existent function. 

This patch advises auto-complete so It waits for tern to obtain its suggestions from the server before popping up. 

-----Original Message-----

kiwanami:
Hi, 
I'll check this code.

kiwanami:
I have checked. The action for synchronous command moves good.

In this PR, the default value of customize variable `tern-ac-sync` is `t`.
Though this code changes completion behavior, I think almost users feel this interaction reasonable.

One point, at the advice `auto-complete`, some major mode symbols are hard coded.
I think it should be extracted as a customize (or just defvar) variable to maintenance easily, because the popular major mode would change in the future. 

Then, we should merge this code!

renatofdds:
Thank you @kiwanami.
I agree with extracting the major modes into a defcustom.

marijnh:
Merged as 4f24418b and followed up with fba79094ca15

PR #548, opened 2015-05-29
Add angular $q constructor type info

KylePDavis:
A minor bug fix to add the `$q` "constructor" type info as described here:
* http://docs.angularjs.org/api/ng.$q

This prevents erroneous complaints about the misuse of `$q` when using the `angular` and `lint` plugins.


marijnh:
Thanks! Merged as 1ac88f5

PR #549, opened 2015-05-29
fix typo in angular $q.reject type info

KylePDavis:


marijnh:
Great. Merged as 5041cfe

PR #550, opened 2015-05-29
Provide a preParse event (to support HTML files)

angelozerr:
This PR provides the new `preParse` event which is called before parsing the file.

I'm developping a new tern plugin which is able to :

 * extract JavaScript content from HTML file.
 * provides completion for HTML ids to support that: 

```html
<html>
  <input id="MyId"  />
  <script>
    var input = document.getElementById(" // here Ctrl+Space shows MyId
    input.ch // here Ctrl+Space shows checked because input is an HTML input element
  </script>
</html>
```

### DIFF
--- a/doc/src/manual.txt
+++ b/doc/src/manual.txt
@@ -815,7 +815,7 @@ object.
 To push code into a context, you first parse it, yielding a syntax
 tree, and then tell Tern to analyze that.
 
-[[infer.parse]]`infer.parse(text: string) → AST`:: Parse a piece of
+[[infer.parse]]`infer.parse(file: File, text: string, passes: [], options?: {}) → AST`:: Parse a piece of
 code for use by Tern. Will automatically fall back to the
 error-tolerant parser if the regular parser can't parse the code.
 
--- a/doc/src/manual.txt
+++ b/doc/src/manual.txt
@@ -815,7 +815,7 @@ object.
 To push code into a context, you first parse it, yielding a syntax
 tree, and then tell Tern to analyze that.
 
-[[infer.parse]]`infer.parse(file: File, text: string, passes: [], options?: {}) → AST`:: Parse a piece of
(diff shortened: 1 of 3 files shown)


angelozerr:
@marijnh here a screenshot of compeltion of element ids that I start managing : 

getelementbyidcompletion

Hope you will like it and you will accept this PR.

marijnh:
Do you need the `partial` flag in your implementation? How are you using it? `infer.parse` already has a mechanism for running passes, it would be nice if we could call `preParse` in the same place as `postParse`.

angelozerr:
At first here my use of preParse https://github.com/angelozerr/tern-browser/blob/master/browser.js#L41 This preParse load with sax the given text (HTML which contains script elements), extract the JavaScript (and replace other characters by spaces) and collect attributes ids which is stored in the give file in the new dom field.

> Do you need the partial flag in your implementation? 

I did that because I told me that dom should be built only for no partial, but to be honnest with you, I have never used partial.

> infer.parse already has a mechanism for running passes, it would be nice if we could call preParse in the same place as postParse.

I agree with you, but parse method has (text, passes, options) parameters signature. To support preParse, file must be added in the signature (file, text, passes, options). If you are OK with that, I will do create a new PR.

angelozerr:
@marijnh I have created the project https://github.com/angelozerr/tern-browser-extension where you can see provided features. Hope you will like it and you will accept this PR. Thanks!

Here the result : 

CodeMirror & GetElementById completion

marijnh:
Changing the signature of `parse` is definitely not ok. The File object is also present in the options, maybe just pass the options object to the pass and have your code access `options.directSourceFile`.

angelozerr:
@marijnh I have removed file parameter and pass teh optiosn to the preParse method. Hope you will be OK with that.

marijnh:
Thanks. Merged as aa14fe7

angelozerr:
thanks @marijnh! https://github.com/angelozerr/tern-browser-extension can now work with the official ternjs. Very cool!

PR #551, opened 2015-05-30
support top scope references prefixed with a '$'

bennidhamma:
This approach works with a .tern-project file that looks like this:

```
{
  "libs": [
    "browser",
    "jquery",
    "ScoutData"
  ],
  "plugins": {
    "node": {
      "modules": {
        "ScoutData": "$ScoutData"
      }
    }
  }
}
```

Basically, it scans the context topScope to see if the module value (prefixed with a '$') exists and uses that if it can.

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -99,8 +99,8 @@
       result = data.modules[name];
     } else if (data.options.modules && data.options.modules.hasOwnProperty(name)) {
       var mod = data.options.modules[name];
-      if (typeof(mod) == "string" && mod.charAt(0) == '$') {
-        result = cx.topScope.props[mod.substring(1)];
+      if (typeof(mod) == "string" && mod.charAt(0) == '=') {
+        result = infer.def.parsePath(mod.slice(1));
       } else {
         var scope = buildWrappingScope(cx.topScope, name);
         infer.def.load(data.options.modules[name], scope);
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -98,9 +98,14 @@
     } else if (name in data.modules) {
       result = data.modules[name];
     } else if (data.options.modules && data.options.modules.hasOwnProperty(name)) {
-      var scope = buildWrappingScope(cx.topScope, name);
-      infer.def.load(data.options.modules[name], scope);
-      result = data.modules[name] = scope.exports;
+      var mod = data.options.modules[name];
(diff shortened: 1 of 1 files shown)


bennidhamma:
To see it in action, clone this repo: https://github.com/bennidhamma/tern-test-node-plugin.

plugin/node.js
```diff
@@ -129,7 +134,12 @@
   }
 
   function postLoadDef(data) {
-    var cx = infer.cx(), mods = cx.definitions[data["!name"]]["!node"];
+    var cx = infer.cx();
+    var def = cx.definitions[data["!name"]];
+    if (!def || !def["!node"]) {
+      return;
+    }
+    var mods = def["!node"];
```
marijnh: What are these changes for?

plugin/node.js
```diff
@@ -98,9 +98,14 @@
     } else if (name in data.modules) {
       result = data.modules[name];
     } else if (data.options.modules && data.options.modules.hasOwnProperty(name)) {
-      var scope = buildWrappingScope(cx.topScope, name);
-      infer.def.load(data.options.modules[name], scope);
-      result = data.modules[name] = scope.exports;
+      var mod = data.options.modules[name];
+      if (typeof(mod) == "string" && mod.charAt(0) == '$') {
+        result = cx.topScope.props[mod.substring(1)];
```
marijnh: This means you can only directly load globals. See [this part](https://github.com/marijnh/tern/blob/master/plugin/requirejs.js#L72) of the requirejs plugin, which uses `infer.def.parsePath` instead, so that it can refer to arbitrary paths. Also, it would be good to stick to the convention used in that plugin and use `=` as the leading character to denote references to the existing environment.

plugin/node.js
bennidhamma: just for debugging. It was hard to step through these lines in the node debugger. I can revert them if you like.

plugin/node.js
bennidhamma: That makes good sense. I'll try and get that working. And I'll switch it over to the equal character. Thanks marin.

bennidhamma:
PTAL (do we say that here?)

marijnh:
Well, I still had to revert the debugging changes, but other than that it looks good. Merged as 9b278a2

PR #555, opened 2015-06-04
Multiline JsDoc

linde12:
With this line-breaks in jsdoc's will be kept. The description also(and if this behavior is unwanted i'll submit a patch) gets cut where the user adds a line-break so it becomes more readable.

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -110,6 +110,36 @@
 
   // COMMENT INTERPRETATION
 
+  function stripLeadingChars(lines) {
+    for (var head, i = 1; i < lines.length; i++) {
+      var line = lines[i], lineHead = line.match(/^\W*/)[0];
+      if (lineHead != line) {
+        if (head == null) {
+          head = lineHead;
+        } else {
+          var same = 0;
+          while (same < head.length && head.charCodeAt(same) == lineHead.charCodeAt(same)) ++same;
+          if (same < head.length) head = head.slice(0, same)
+        }
+      }
+    }
+    lines = lines.map(function(line, i) {
+      line = line.replace(/\s+$/, "");
+      if (i == 0 && head != null) {
+        for (var j = 0; j < head.length; j++) {
+          var found = line.indexOf(head.slice(j));
+          if (found == 0) return line.slice(head.length - j);
+        }
+      }
+      if (head == null || i == 0) return line.replace(/^[\s\*]*/, "");
+      if (line.length < head.length) return "";
+      return line.slice(head.length);
+    });
+    while (lines.length && !lines[lines.length - 1]) lines.pop();
+    while (lines.length && !lines[0]) lines.shift();
+    return lines;
+  }
+
   function interpretComments(node, comments, scope, aval, type) {
     jsdocInterpretComments(node, scope, aval, comments);
     var cx = infer.cx();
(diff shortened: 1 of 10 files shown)


linde12:
I see a lone break test has failed. But as i said before, it is in my opinion better to be able to put line breaks where you want in your JsDoc description. The test is probably failing because there a string with line breaks will be returned(which it has to do if we're going to put each @-attribute on a new line anyway) so that test will have to be changed anyway.

linde12:
I'm still a bit confused how the testing works - everything's completely new for me.

@marijnh please tell me what you think about the line breaks in the JsDocs. (Now not only for @-attributes, but also for the description). I think it's an improvement.

linde12:
Later maybe we could add some parsing for things like `@param`, `@author`, `@return`, etc and make headings for each section where we list the different properties like this:
```
addMessage(message: {alias: ?, text: ?})

Description
    Dispatches a message
Params
    message(Object) - Object containing information about the message
    ....
Return
    Number
```
Would be really useful when working with large apps that need easily accessible documentation.

linde12:
I think it would be good to run the test suite with and without fullDocs and check if all * are stripped away. Last fix commit was due to me not seeing fullDocs. However i'm not entirely sure how the tests work, so i'm probably not good for it.

plugin/doc_comment.js
```diff
@@ -123,12 +123,15 @@
     var result = comments[comments.length - 1];
     if (cx.parent._docComment.fullDocs) {
       result = result.trim().replace(/\n[ \t]*\* ?/g, "\n");
+      result = result.replace(/^\*/g, "");
     } else {
       var dot = result.search(/\.\s/);
       if (dot > 5) result = result.slice(0, dot + 1);
-      result = result.trim().replace(/\s*\n\s*\*\s*|\s{1,}/g, " ");
+      // Remove all space*space in beginning of the lines
+      result = result.replace(/^\n?([ \t]*\* *)/gm, "");
```
marijnh: This is going to match at most one time (since it includes `^`). That seems unlikely to be what you intended.

marijnh: Ah, I see, you're adding an `m` flag.

linde12:
Yeah, i've been running it at work now for a couple of days and i've had no problems so far. I also added "departments" in the Sublime plugin like this:

getProduct(something: String) -> ...

**Description**
Some description here for people to read

**Parameters**
{String} something desc

**Returns**
{Object} abc

**Author**
linde12

but i'm not happy enough with it to push it just yet - it works but needs refinement.

marijnh:
I've thought about this some more, my first impression being that this should be configurable, but in fact, there are various ways in which client plugins might want to consume comments, and trying to encode them all into options makes for too much complexity, and a lot of awkwardness to ensure the options end up being right.

So I've removed `fullDocs` in the attached comment, and made the doc_comments plugin always return the whole comment, with newlines. I expect this'll cause some awkwardness in some plugins, but we'll work it out -- they can strip or clip the strings on the client side.

I've also written an indentation-stripper that I think works better, but do test it with your comment style to see whether it holds up.

linde12:
I totoally agree regarding the fullDocs, keep it simple! And as you said they can manipulate it on the client.

I'll try the new comment stripping once i'm home! :+1:

linde12:
It works just like before! Great!

marijnh:
Hey @linde12, a heads-up: the new default was causing problems for people, so I've now added a _query_ option, rather than a plugin option, that you have to opt-in to to get full comment strings. Add `docFormat: "full"` to your queries.

linde12:
Ah, sorry for the late response. That seems reasonable!

PR #560, opened 2015-06-11
Completion for required/defined modules

angelozerr:
This PR provides completion for RequireJS modules inside string array dependencies (which looks like to node tern plugin).

Perhaps it should be improved by using baseurl, but it's the first step for RequireJS completion module.


angelozerr:
I close this PR because of merge conflict. See new PR https://github.com/marijnh/tern/pull/573

PR #438, opened 2014-12-09
#424 plugins.json generated from metadata in tern.java

paulvi:
plugins.json generated from metadata in tern.java

29 entries

tern-plugins

see discussion in https://github.com/angelozerr/tern.java/pull/190

PR #561, opened 2015-06-12
Basic support for JSDoc @param {function}

angelozerr:
This PR provides a basic support for JSDoc {function} : 

```javascript
/**
 * @param {function} fn
**/
function test(fn) {
}
```

It should be improved to :

 * support args of function : 

```javascript
/**
 * @param {function(number)} fn
**/
function test(fn) {
}
```

 * support this of function : 

```javascript
/**
 * @param {function(this:number)} fn
**/
function test(fn) {
  this. // here Ctrl+Spaces shows methods of number.
}
```

In other words it should be cool to have the same features than JetBrains. See http://blog.jetbrains.com/webide/2012/10/validating-javascript-code-with-jsdoc-types-annotations/

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -304,6 +304,9 @@
           if (str.charAt(pos++) != ">") return null;
           val.type.propagate(type.defProp("<i>"));
         }
+      } else if(/^function$/i.test(word)) {
+        // TODO : support for args parameter types + return types + "this"
+        type = new infer.Fn(null, infer.ANull, [], [], infer.ANull);
       } else {
         while (str.charCodeAt(pos) == 46 ||
                acorn.isIdentifierChar(str.charCodeAt(pos))) ++pos;


marijnh:
Is this specified somewhere? I can't find anything about it on in the JSDoc docs.

angelozerr:
@marijnh  you are right, it seems that JSDoc doens't support that. @callback is used for that. See : 

 * https://github.com/jsdoc3/jsdoc/issues/260
 * http://usejsdoc.org/tags-callback.html

Perhaps it could be interesting to gives the capability to extends doc_comment to support a custom syntax?

marijnh:
Wow, that's a strange community. They seem entirely unable to deal with recursive syntax, or to grasp the fact that typedefs could already play the part of defining function types, or to notice that _callback_ isn't a proper name for all function types.

So, um, when something is finally decided on, I guess we could support it. I'd also be okay with supporting the much saner Google Closure syntax. But I'd want an actually useful implementation that also parses parameter and return types -- just saying that something is a function can already be done by using `Function` as its type, I'm not sure this patch adds anything.

angelozerr:
Yes you are right we can use `@param {Function}` instead of using `@param {function}. My patch give sthe capability to use `Function` or `function`.

But my orignal problem is for my tern-lint. I have a validator which validates if a token is a function. To do  that I do `type.getFunctionType()`. In the case when type is a `Function.prototype`, it returns null. I don't know if it's a bug?

marijnh:
I didn't know `Function.prototype` was a function (the only function, apparently, whose prototype is `Object.prototype`). Attached patch fixes this. Does that solve your problem?

paulvi:
:+1: sign up

paulvi:
please comment why just closing in #561 and #438

marijnh:
There was no further response for five months, so I assumed the discussion was over.

PR #567, opened 2015-06-20
Improve RequireJS Error callback JSON Type Definition

angelozerr:
See http://requirejs.org/docs/api.html#errors


marijnh:
Great. Merged as 2d827ad

PR #571, opened 2015-06-26
Promise#then and Promise#catch must returns an instance of Promise

angelozerr:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then


angelozerr:
Perhaps we should return `!this` instead of `+Promise`. I don't know that.

PR #573, opened 2015-06-28
Completion for required/defined modules

angelozerr:
This PR provides completion for RequireJS modules inside string array dependencies (which looks like to node tern plugin).


marijnh:
Merged in 21c5ce41c85f, with some surrounding work to reduce duplication in a3bc3b4718beb and e77317059229a1552

PR #575, opened 2015-07-01
Add url + doc for find type of required module

angelozerr:
This PR add url + doc for find type of required module & fill !doc/!url for required module (fs, etc). here a screenshot where find type is done for "fs" required module : 

nodefindtype


marijnh:
Thanks. Merged as 3d6debe

PR #578, opened 2015-07-08
Remove a repeated word.

happibum:


marijnh:
Thanks. Merged as 1b15073

PR #581, opened 2015-07-10
Set custom metaData to the tern file

angelozerr:
This PR gives the capability to set custom metaData to the tern file when tern file is send to teh server by the client.

For instance tern client can set metaData to {"html": true} and this meta data can be used by a tern plugin to know if the file is an html file.

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -111,10 +111,10 @@
     this.reset();
   };
   Server.prototype = signal.mixin({
-    addFile: function(name, /*optional*/ text, parent) {
+    addFile: function(name, /*optional*/ text, parent, /*optional*/ metaData) {
       // Don't crash when sloppy plugins pass non-existent parent ids
       if (parent && !(parent in this.fileMap)) parent = null;
-      ensureFile(this, name, parent, text);
+      ensureFile(this, name, parent, text, metaData);
     },
     delFile: function(name) {
       var file = this.findFile(name);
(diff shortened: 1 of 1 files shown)


marijnh:
In the case of HTML files, wouldn't the filename already contain that information?

angelozerr:
I need that because i would like to add some option when file is posted like set that file is an html file to support jsp xhtml php extension and customize the script tag which contains js
In my case js can be hosted inside script and aui:script

marijnh:
What I was asking is, can't you look at the extension in the filename to figure out what kind of file it is?

angelozerr:
Yes i have done that but when i extract js from file i need script tags option

I tell me that metadata can be usefull to set custom data from the client

marijnh:
Could we simply have `addFile` return the file object, and your code adding a property to that, so that we don't have to put an extra feature into the core code?

angelozerr:
How to you use that with .tern with node?

marijnh:
I don't understand that last question.

angelozerr:
> I don't understand that last question.

Forget my comment.

In my case I need to :

 * set a custom property with (java) tern client.
 * retrieve this custom property inside preParse https://github.com/marijnh/tern/blob/master/lib/infer.js#L1285 so the addFile solution will not work for me

marijnh:
`preParse` can look at the file name, no? So again, why can't you use that?

PR #590, opened 2015-07-26
Add missing parameters for Array/String/JSON methods

adrianton3:


defs/ecma5.json
```diff
@@ -233,7 +233,7 @@
         "!doc": "Returns the last index at which a given element can be found in the array, or -1 if it is not present. The array is searched backwards, starting at fromIndex."
       },
       "every": {
-        "!type": "fn(test: fn(elt: ?, i: number) -> bool, context?: ?) -> bool",
+        "!type": "fn(test: fn(elt: ?, i: number, array: !this) -> bool, context?: ?) -> bool",
```
marijnh: You can't use `!this` here (you can only introduce terms that depend on the input of a function type in its return type).

You could use `+Array` for the type here and update the effects for these functions to pass `!this` as a third parameter (by adding it to the end of the effect string)

marijnh:
Apologies about completely losing track of this. Merged as 0338aae...87eee294

PR #591, opened 2015-07-27
Fix promise chaining behavior

othree:
Fix the chaining behavior of promise

Sample code:

```js
var pp = new Promise(function (res) {
  res(1);
});
pp.then(function (num) {
  num. // provide number properties
  return '';
}).then(function (str) {
  str // should provide string properties
})
```

### DIFF
--- a/defs/ecma6.json
+++ b/defs/ecma6.json
@@ -10,7 +10,7 @@
       "then": {
         "!doc": "The then() method returns a Promise. It takes two arguments, both are callback functions for the success and failure cases of the Promise.",
         "!url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then",
-        "!type": "fn(onFulfilled: fn(value: ?), onRejected: fn(reason: ?)) -> !this",
+        "!type": "fn(onFulfilled: fn(value: ?), onRejected: fn(reason: ?)) -> !custom:Promise_then",
         "!effects": [
           "call !0 !this.value"
         ]
--- a/lib/def.js
+++ b/lib/def.js
@@ -592,5 +592,21 @@
     return self;
   });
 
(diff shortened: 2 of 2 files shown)


othree:
This will break at the following code:

```js
pp.then(null, function () {}).then(function (res) {
  res. // will always be ? 
});
```

othree:
Maybe we can add new syntax like  `||`, `or`.
And bahaves like JavaScript `||`.
If the first(LHS) is ANull, def.js will keep looking the second one(RHS).

marijnh:
There's also the case of the resolution callback returning another promise. I think that to handle this you'll need to add custom code (there's already `Promise_ctor` in `lib/def.js`, if you want to work on this, you'd define another custom hook there, and have the definition of Promise.then refer to that).

othree:
Ya, I forgot about that case. Will take a deep look.

othree:
@marijnh , May I ask some basic question here (or any other suggestion place)?
I want to fix this. But I am newbie about parser.

marijnh:
Definitely. Be sure to read http://marijnhaverbeke.nl/blog/tern.html as background first.

othree:
Thanks, here is first question,want to confirm I realize the code correctly.

Is https://github.com/marijnh/tern/blob/master/lib/def.js#L591

```js
args[0].propagate(new infer.IsCallee(infer.ANull, [exec, reject], null, infer.ANull));
```

Mapping to https://github.com/marijnh/tern/blob/master/defs/ecma6.json#L486

```
executor: fn(resolve: fn(value: ?), reject: promiseReject)
```

If yes, is `args[0]` means `executor` function?

othree:
Update my path to use a new custom constructor.

marijnh:
Great. I've merged this as c5baa809724 and updated it in c5df4871f109 (to make sure it still works if the return type of the function doesn't become clear until after the custom function runs).

PR #592, opened 2015-07-27
Replaced HTTP references to nodejs.org with HTTPS

jonathanp:
Replaced HTTP references to nodejs.org with HTTPS to avoid server side redirects that loses the hash fragments of the URL's.

For example if I go to [http://nodejs.org/api/os.html#os_os_tmpdir](http://nodejs.org/api/os.html#os_os_tmpdir) I end up at the top of the page while if I go to [https://nodejs.org/api/os.html#os_os_tmpdir](https://nodejs.org/api/os.html#os_os_tmpdir) I get to the correct position.

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -330,109 +330,109 @@
         "!type": "fn(id: string) -> !custom:nodeRequire",
         resolve: {
           "!type": "fn() -> string",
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Merged as f67c634

Issue #556, opened 2015-06-04
`tern-before-change` hook causes jumping when used with `evil-mode` (Emacs)


emallson:
**The Bug:** When I delete a segment of code with `dfx` (delete forward to the first x), the buffer scrolls to several very different positions in a very short period of time. Removing `tern-before-change` from the `before-change-functions` list causes this to stop.

**Reproducing:**

1. Install Emacs, Tern, Tern-Mode and Evil-Mode.
2. Begin editing a JS file. Write out a function, then go to the declaration line (`function foo(bar) { ... `) and run `df{`. The screen will jump upwards and then back down very rapidly (if the buffer cannot be scrolled then it does not occur).

Changing the "200 millisec" to  "600 millisec" in `tern-before-change` fixes it, but I am not sure what effect that would have on the actual behavior yet.

Also: it seems that `tern-before-change` is actually trying to wait until *after* the change to send the buffer, so why not use `after-change-functions`? I will experiment with this after work. Wanted to get it reported so I wouldn't forget.

emallson:
I believe I have fixed this (without breaking anything) in https://github.com/emallson/tern/commit/6f62857086e7bb16a33f04e3d7a9a8ae544c5493, but I'm going to give it a few days of usage before submitting a PR because there don't seem to be any elisp tests.

marijnh:
Has the patch stood the test? Should I merge it?

emallson:
Not quite. Emacs idle-timer functions are not buffer-local as I'd initially thought. The result is that `tern-reparse-on-idle` gets run on other buffers. I need to fix that. Aside from that, it seems to work quite well.

emallson:
A quick update on this: sorry that it took me so long. I was working with ES6/7 code for a while and couldn't really test my change on that. I've been back on ES5 code and the change I made has worked great. I'll submit a PR this evening once I'm off work.

marijnh:
(Still hoping for that PR!)

emallson:
@marijnh: sorry! just submitted PR

marijnh:
Done via #593

PR #593, opened 2015-07-27
[Emacs] Use `after-change-hook` instead of `before-change-hook`. 

emallson:
**Summary of Changes**

- Use an `after-change-hook` for tracking dirty region of buffer instead of `before-change-hook` + a timeout. This is more in line with the intended use of those hooks.
- Use an idle timer to submit updated files to the server instead of running that in the change hook. This prevents the submission of large files from interfering with normal editing. The default timer is 1s of non-activity.

I've been using this for a couple of weeks now (installed via el-get) and haven't had any issues with it. I'd appreciate it if a couple of other emacs users would try it out and report back. Going to ask on `#emacs` today.

Resolves issue #556


emacs/tern.el
```diff
@@ -510,24 +512,28 @@ list of strings, giving the binary name and arguments.")
...
-  (setf tern-last-point-pos nil)
-  (when (and tern-last-argument-hints (<= start (car tern-last-argument-hints)))
-    (setf tern-last-argument-hints nil))
-  (when (and tern-last-completions (<= start (cadr tern-last-completions)))
-    (setf tern-last-completions nil)))
+      (setf tern-buffer-is-dirty (cons (min start (car tern-buffer-is-dirty))
+                                       (max end   (cdr tern-buffer-is-dirty))))
+    (setf tern-buffer-is-dirty (cons start end))))
+
+(defvar tern-idle-time 1
```
marijnh: That is 1 second, right? Why so low? Even during regular typing, you might get a one-second pause, and since Tern might do something mildly expensive, I'd be tempted to set this higher (5 seconds?)

emallson: It is user configurable (`C-h v tern-idle-time RET`). 1 second has worked well for me, but YMMV. IMO 5 seconds would be too high, but I can see how 1 second would be too low. Maybe meet in the middle and do 2.5?

marijnh:
Great. Merged as ae3a3e7, and moved to a higher delay time in 48b59d9

PR #595, opened 2015-08-02
[requirejs] ignore current module for required module completion

angelozerr:
This PR ignore the current module  for required module completion: js extension must be removed to retrieve module name.


marijnh:
Great. Merged as 3454990

PR #596, opened 2015-08-02
Support for doc & url for required module

angelozerr:
This PR looks like findTypeAt of node tern plugin, to support doc and url when find type is done for required module.

### DIFF
--- a/plugin/requirejs.js
+++ b/plugin/requirejs.js
@@ -255,19 +255,26 @@
   });
 
   function findTypeAt(_file, _pos, expr, type) {
-    if (!expr || expr.node.type != "Literal" ||
-        typeof expr.node.value != "string" || !expr.node.required)
-      return type;
+    if (!expr) return type;
+    var isStringLiteral = expr.node.type === "Literal" &&
+       typeof expr.node.value === "string";
+    var isRequireArg = !!expr.node.required;
 
-    // The `type` is a value shared for all string literals.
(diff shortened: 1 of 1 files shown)


plugin/requirejs.js
```diff
@@ -255,19 +255,26 @@
...
-    // Provide a custom origin location pointing to the require()d file
-    var exportedType = expr.node.required;
-    type.origin = exportedType.origin;
-    type.originNode = exportedType.originNode;
+    if (!expr) return type;
+    var isStringLiteral = expr.node.type === "Literal" &&
+       typeof expr.node.value === "string";
+    var isRequireArg = !!expr.node.required;
+
+    if (isStringLiteral && isRequireArg) {
```
marijnh: Why the apparently random refactor to use boolean variables and an extra condition rather than the existing early return?

angelozerr: I have copied/pasted the code from node tern plugin and adapt it for requirejs. If you don't like this code, I think we should modify node tern plugin too.

mdasrafulalamasa:
My resume and the first

plugin/requirejs.js
angelozerr: @marijnh do you want a new PR for that?

marijnh:
Does attached two-line patch do what you needed?

angelozerr:
Yes thanks @marijnh. I close this PR.

PR #600, opened 2015-08-08
Fix grammar: An degradation -> A degradation

cdosborn:


marijnh:
Thanks. Merged as 82189ae

PR #602, opened 2015-08-12
[node] Add missing methods for Writable + use multiple type for chunk

angelozerr:


paulvi:
Otherwise there's error just for Node.js Hello World

angelozerr:
When node tern plugin was written, tern didn't support multiple parameter. Today tern supports multiple parameter and now tern-lint too. This PR fixes the problem of @paulvi but I think there are other problems. Any contribution are welcome for tern node plugin to update JSON Type Definition with multi type.

marijnh:
Thanks. Merged as 7bf4d8d

PR #603, opened 2015-08-13
Support find type for propety key of object literal

angelozerr:
This PR support find type for propety key of object literal. Here a screenshot of find type of RequireJS config of callback property key (when  callback property key is hoverd with mouse inside Eclipse): 

findtyperequirejsconfig

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -791,7 +791,7 @@
         (objProp = pointInProp(expr.node, resolvePos(file, query.end)))) ||
         (expr.node.type == "Property" && (objProp = expr.node))) {
       var name = objProp.key.name;
-      if (expr.node.type == "Property") expr = {node: exports.parentNode(objProp, file.ast), state: expr.state};
+      if (expr.node.type == "Property") expr = {node: infer.parentNode(objProp, file.ast), state: expr.state};
       var fromCx = ensureObj(infer.typeFromContext(file.ast, expr));
       if (fromCx && fromCx.hasProp(name)) {
         type = fromCx.hasProp(name);
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -749,8 +749,14 @@
       if (expr) return expr;
       expr = infer.findExpressionAround(file.ast, start, end, file.scope);
       if (expr && (expr.node.type == "ObjectExpression" || wide ||
-                   (start == null ? end : start) - expr.node.start < 20 || expr.node.end - end < 20))
+                   (start == null ? end : start) - expr.node.start < 20 || expr.node.end - end < 20)) {
+        if (expr.node.type == "ObjectExpression" ) {
+          // return object literal property if possible.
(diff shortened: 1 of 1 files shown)


lib/tern.js
```diff
@@ -749,8 +749,14 @@
       if (expr) return expr;
       expr = infer.findExpressionAround(file.ast, start, end, file.scope);
       if (expr && (expr.node.type == "ObjectExpression" || wide ||
-                   (start == null ? end : start) - expr.node.start < 20 || expr.node.end - end < 20))
+                   (start == null ? end : start) - expr.node.start < 20 || expr.node.end - end < 20)) {
+        if (expr.node.type == "ObjectExpression" ) {
+          // return object literal property if possible.
+          var objProp = pointInProp(expr.node, end);
+          if (objProp) return {node: objProp, state: expr.state};
```
marijnh: I don't think it's a good idea for `findExpr` to return a non-expression (as your special casing further down in the patch shows, it causes problems). Would it make sense to move this test (whether the expr is an object's property) into `findTypeAt` instead?

angelozerr: I have created this PR to try to implement find type for property key, but it seems that it's not the best mean since it breaks tests. I hope you will accept this feature by implementing it with a clean mean.

marijnh:
Looking at the code, it seems this already works -- fetching the type of an object literal problem does yield the intended type. What is this pull request trying to accomplish, then?

angelozerr:
> What is this pull request trying to accomplish, then?

See screenshot. You can hover on the property key.

marijnh:
What kind of request does hovering make in your system?

angelozerr:
> What kind of request does hovering make in your system?

I use find type (like Ctrl+i with your CodeMirror demo).

marijnh:
Okay, and that does give an accurate type in my demo. So once again, please define the actual problem.

angelozerr:
>  So once again, please define the actual problem.

Ok sorry. The problem is about documentation. If you see my screenshot, the hover displays the documentation of the callback function and not teh documentation about the function type.

marijnh:
I'm getting the correct documentation when I do ctrl-i in the demo.

angelozerr:
You are right, it works for function type, but not for simple type like string, bool, etc. Ex : 

```javascript
require.config({
  baseUrl
})
```

marijnh:
I think attached patch will help.

angelozerr:
Works like a charm, and the last case which doesn't work is : 

```javascript
var c = {baseUrl // here doc of baseUrl doesn't work
require.config(c)
```

marijnh:
This is just how type queries work -- they don't have the fallback magic that completion has, and will act on the concrete type of the thing you are querying, if it has one. In this case, `c` simply has type `{baseUrl: ?}`.

angelozerr:
> This is just how type queries work

Ok I understand, This case does not happen often.

PR #604, opened 2015-08-14
Completion & Find type for angular dependencies module.

angelozerr:
This PR supports 

 * completion for angular dependencies module:

modulecompeltion

 * find type for angular dependencies module:

modulefindtype

@marijnh I have improved a lot angular tern plugin with my fork. So I would like to contribute with my forked angular tern plugin :  my idea is to create little PR, and when you accept it, I will create other PR. Hope you will like this idea.

### DIFF
--- a/plugin/angular.js
+++ b/plugin/angular.js
@@ -128,10 +128,11 @@
     if (node && node.type == "ArrayExpression")
       for (var i = 0; i < node.elements.length; ++i) {
         var elt = node.elements[i];
-        if (elt.type == "Literal" && typeof elt.value == "string")
+        if (elt.type == "Literal" && typeof elt.value == "string") {
           // mark node as module.
           elt.module = true;
           strings.push(elt.value);
+        }
       }
     return strings;
   }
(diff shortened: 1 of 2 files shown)


marijnh:
The `findTypeAt` functionality was already accepted in June, so I'm not sure why it is included again here. Could you try to port this to the current code and submit a cleaned-up pull request?

angelozerr:
@marijnh I will be happy to create a new patch, but if you see I mark the node as module https://github.com/marijnh/tern/pull/604/files#diff-ffad115ba87519a141dc5c55a7590a62R133 so I'm not sure that you like that.

marijnh:
Maybe create a predicate that checks for the proper parent nodes instead?

angelozerr:
> Maybe create a predicate that checks for the proper parent nodes instead?

Yes I will try do that, but I don't understand why you don't like? It's because it's not clean?

marijnh:
If all modules start putting properties on some nodes (but not others), we'll end up a lot of different object shapes, which will prevent the JIT from compiling the library efficiently.

PR #613, opened 2015-08-27
[node] Simplify completion for required module

angelozerr:
When node string is inside the require function, it is marked with node.required. So we can just check this mark to know if completion is inside require function.

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -208,15 +208,9 @@
   // Completes CommonJS module names in strings passed to require
   function findCompletions(file, query) {
     var wordEnd = tern.resolvePos(file, query.end);
-    var callExpr = infer.findExpressionAround(file.ast, null, wordEnd, file.scope, "CallExpression");
-    if (!callExpr) return;
-    var callNode = callExpr.node;
-    if (callNode.callee.type != "Identifier" || callNode.callee.name != "require" ||
-        callNode.arguments.length < 1) return;
-    var argNode = callNode.arguments[0];
-    if (argNode.type != "Literal" || typeof argNode.value != "string" ||
-        argNode.start > wordEnd || argNode.end < wordEnd) return;
-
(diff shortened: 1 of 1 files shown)


marijnh:
This has already been rewritten as part of the module loader overhaul.

angelozerr:
Yes I have seen that. I have written a lot of tern modules (for YUI, tabris, browser extension, etc) and I think more and more that the best mean is to mark the node with a module (or other thing) instead of sexecute expressionType inside custom completion :

 * before my PR, teh custom completion check if node was inside require function by using CallExpression
 * after my PR you need not that, you use teh marked node (node.required).

When you mark the node with something, you can use it:

 * for custom typeAt
 * for custom completion
 * for validation (with my tern-lint)

without searching where you are with infer.findExpressionAround like you are doing for requirejs tern plugin (I will do a PR to change that).

Hope you will understand my comment.

marijnh:
I removed the `required` field for a reason -- adding random fields to existing objects isn't great, and if we can do it with a predicate function, like I am doing now, I prefer that. So don't create a PR unless you have a really good reason to.

angelozerr:
> So don't create a PR unless you have a really good reason to. The well reason is that you can use :

 * for custom typeAt
 * for custom completion
 * for validation (with my tern-lint)

For instance, requirejs uses node.required 

 * for custom typeAt : https://github.com/marijnh/tern/blob/master/plugin/requirejs.js#L278
 * it could use it for custom completion https://github.com/marijnh/tern/blob/master/plugin/requirejs.js#L285 to avoid calling infer.findExpressionAround (and simplify code like I have done in this PR)
 * I use it for validation https://github.com/angelozerr/tern-requirejs-extension

marijnh:
You may be able to use the `isModName` and `isImport` methods on the `modules` plugin object (`server.mod.modules`) for this purpose (though you'll still need to resolve the resulting name -- see the logic in, for example, `findTypeAt` in that plugin.

PR #614, opened 2015-08-28
[node] Use resolveProjectPath + resolveModulePath in an another plugin

angelozerr:
I'm developping the capability to display list of node modules once the user apply completion https://github.com/angelozerr/tern-guess-types

For that I need to create my own list of modules like https://github.com/marijnh/tern/blob/master/plugin/node.js#L247. So I need resolveProjectPath + resolveModulePath and to avoid copying/pasting your code in lmy tern plugin, I would like to expose it.

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -176,7 +176,9 @@
       currentFile: null,
       currentRequires: [],
       currentOrigin: null,
-      server: server
+      server: server,
+      resolveProjectPath: resolveProjectPath,
+      resolveModulePath: resolveModulePath
     };
 
     server.on("beforeLoad", function(file) {


marijnh:
Hi Angelo,

The ES6 work heavily refactored the module loader. You can now implement your own resolution strategies (see the `resolvers` property on the `modules` plugin), which might already give you the information you're looking for. Also be aware that file names are now resolved to project-relative paths, rather than absolute paths, which might require some changes in your code.

Please take a look at the new code, and let me know whether it works as-is, or whether you need additional interfaces.

angelozerr:
Thanks @marijnh ! I will give you feedback.

angelozerr:
@marijnh if you accept this PR https://github.com/marijnh/tern/pull/621 it will be enough for me.

marijnh:
Great. I've merged #621

PR #616, opened 2015-08-29
Added the Image() constructor

dani-h:
The Image constructor is an alternative to document.createElement('img')
which returns a HTMLImageElement.


marijnh:
Thanks! Merged as 9073271

PR #621, opened 2015-08-31
[node] Provide getModType

angelozerr:
This PR provides teh capability to use getModType from the given node. It is used here for typeAt, but it could be used by another plugin. In my case I would like to use it to validate the required node module.


marijnh:
Looks great. Merged as 704277e

angelozerr:
Thanks!

PR #622, opened 2015-09-01
Add test for property keys completion inside referenced object literal

angelozerr:


angelozerr:
@marijnh I don't understand, test failed although it works with the demo?

marijnh:
Well, the first problem is that you didn't bother running the tests before opening a PR. I'd really appreciate if you'd started doing that.

The test case runner disables guessing by default, and the completion magic tested here is only enabled when guessing is on. You have to add a question mark to the annotation comment to enable it. I've done this and pushed this as 7d873c7

angelozerr:
> Well, the first problem is that you didn't bother running the tests before opening a PR. I'd really appreciate if you'd started doing that.

Yes I know, but I didn't know why I had this problem, so I hav ecreated a PR. Next time, I will post you the question in an issue.

Thanks for your explanation and fix.

Issue #626, opened 2015-09-02
[node] Ignore current file name for node completion (regression)


angelozerr:
The current file name for node completion should be ignored. Today it is not ignored (before tern-es6, it was ignored)

```javascript
require("./ // here Ctrl+Space shows local files but it should ignore "./a" (current file name)
```

PR #627, opened 2015-09-02
Fix https://github.com/marijnh/tern/issues/626 with Windows OS

angelozerr:
This PR fixes https://github.com/marijnh/tern/issues/626 with Windows OS.

```javascript
var projectPath = path.relative(pDir, path.resolve(dir, file))
```

For instance projectPath is equals to  "js\a.js" and is compared to parentFile  "js/a.js" (the test projectPath  == parentFile) returns false)

This PR replace the \ by /

### DIFF
--- a/plugin/modules.js
+++ b/plugin/modules.js
@@ -197,7 +197,7 @@
   }())
 
   function normPath(name) { 
-    return name.replace(/\\/g, "/")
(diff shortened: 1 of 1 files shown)


marijnh:
I've attached an alternative patch (`parentFile` should already be normalized, and we should normalize the value in `projectPath` since it is also used as a key into `modules`)

PR #629, opened 2015-09-03
Provide basic support for ECMAScript7

angelozerr:
This PR provides a basic `ECMAScript7` : 

 * Initialize `ecma7.json` with new methods `Array.prototype.includes` & `ArrayBuffer.transfer`
 * load the def ecma7 if ecmaVersion is set to 7.


marijnh:
I'd prefer if this was maintained outside of the repository until the standard stabilizes. I don't have time to deal with the churn and changes that will without a doubt happen during the standardization process.

angelozerr:
Ok we will wait the standardization of ES7 do that.

PR #630, opened 2015-09-03
Support for several signature of jQuery.fn.html

angelozerr:
See http://api.jquery.com/html/ to see that html can have several signatures.


marijnh:
Thanks. Merged as d91740f

PR #632, opened 2015-09-11
console.*(text) can wait string or object

angelozerr:
The console.log, etc can wait text or object argument.


marijnh:
Thanks, merged as ca53c4a

PR #633, opened 2015-09-11
Completion for module specifiers

angelozerr:
This PR provides completion for module specifiers. It works only if you type one character in the specifier. Ex : 

```javascript
import {L // here Ctrl+space shows List} from "./list"
```

The completion returns `isSpecifier: true` In my case (tern.java), I need that because my completion provides "List" and "List(head, tail)" (it's helpfull for function, but not for specifier).

### DIFF
--- a/plugin/modules.js
+++ b/plugin/modules.js
@@ -131,6 +131,16 @@
 
       var modType = this.resolveModule(modName, node.sourceFile.name)
       return (prop ? modType.getProp(prop) : modType).getType()
+    },
+    
+    completeSpecifierName: function(modNode, completions, query, word) {
+      function gather(prop, obj, depth) {
+        if (query.filter !== false && word &&
+            (query.caseInsensitive ? prop.toLowerCase() : prop).indexOf(word) !== 0) return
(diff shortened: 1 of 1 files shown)


marijnh:
Code related to `import` declarations does _not_ belong in the modules plugin, which is a generic piece of infrastructure for all kinds of module systems. I've pushed an alternative patch that uses the existing `isImport` functionality to do this in a module-system-independent way.

angelozerr:
Thanks @marijnh for your patch whcih works well and thanks to have keeping isSpecifier: true.

I finnd it's just shame that it doesn't work when no word are typed (I have found a solution to do that).

marijnh:
I looked at your code and didn't see anything to handle this situation. How did you detect that the cursor is in such a position?

angelozerr:
Yes I know I had the same problem than you, I would like just tell you about this limitation and I hoped that you will find a solution, but I'm afraid that it is not possible.

marijnh:
You could do your own custom superficial parse of the source text to figure out whether the position is between braces in an import, but for now I think that's too much complexity.

PR #640, opened 2015-09-18
Add from start/end of the node where find definition is executed

angelozerr:
In Eclipse Tern IDE, when user wish to find definition, it uses Ctrl+Click on a node which call findDef (it is called Hyperlink). I need to underline this node and this PR gives me the capability to find the start/end of the node to underline.

This PR is simple, but after it should be cool to improve it to customize the from start/end like typeAt.

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -922,7 +922,7 @@
     if (infer.didGuess()) return {};
 
     var span = getSpan(type);
-    var result = {url: type.url, doc: parseDoc(query, type.doc), origin: type.origin};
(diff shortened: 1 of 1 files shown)


marijnh:
This seems out of place. Wouldn't it be cleaner to write your own little plugin that adds a new query type that gives you the extent of the expression at a given position?

angelozerr:
> Wouldn't it be cleaner to write your own little plugin that adds a new query type that gives you the extent of the expression at a given position?

I think it's shame to write a tern plugin just to retrieve those information. Perhaps we could add an option to the query to support that? What do you think about that?

marijnh:
No, same story, I don't want to clutter up general-purpose code to add special-purpose functionality if it can be avoided. The plugin could be written in a few lines.

Issue #641, opened 2015-09-20
How to contribute to node module like "!node" in tern 0.13.0


angelozerr:
In https://github.com/marijnh/tern/blob/0.13.0/plugin/node.js#L137 we could contribute to node module with "!node" where tern plugin like https://github.com/angelozerr/tern-node-express was based.

With 0.15.0, it seems "!node" was removed, so my question how to do  contribute to node module like "!node" with tern 0.15.0?

angelozerr:
Ok, I have changed !node with !modules, now my express node modules is known, but compeltion like require('express'). doesn't work.

It seems that the problem is because of node_resolve. My 'express" node modules must be considered as local, but it is not local and https://github.com/marijnh/tern/blob/master/plugin/node_resolve.js#L12 return null.

marijnh:
The rewrite of the module system removed the ambiguity between file names and named modules. The `!modules` field's properties now refer to file names, so you can dump the result of condensing a module into a definition and have it be used for the proper module. Attached patch adds support for a `!known_modules` definition field (under `!define` again), which should work the way you want it to -- you can require the precise name given to a module in this field, and get the definitions under that name.

angelozerr:
It works like a charm thanks! Could you please update the manual too?

marijnh:
I don't believe any of this is in the manual at all. Is it?

angelozerr:
> I don't believe any of this is in the manual at all. Is it?

No, when I said update, I said add a new section in the node plugin. If you wish to contribute to tern node plugin, you can use !known_modules and you could add samples like:

 * https://github.com/angelozerr/tern-node-express
 * https://github.com/angelozerr/tern-node-mongodb-native
 * https://github.com/angelozerr/tern-node-mongoose

PR #642, opened 2015-09-20
Fix https://github.com/marijnh/tern/issues/641

angelozerr:
This PR fix issue https://github.com/marijnh/tern/issues/641 when a node modules is contributed with !modules like https://github.com/angelozerr/tern-node-express


marijnh:
See my attached patch in #641

Issue #649, opened 2015-09-26
throw error when type var a 


agileago:
```
atom-ternjs
TypeError: Cannot read property 'type' of null at Array.isImport
(/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/plugin/commonjs.js:57:13) at
Object.signal.mixin.findIn (/Users/jianhuaren/.atom/packages/atom-
ternjs/node_modules/tern/plugin/modules.js:73:28) at Object.signal.mixin.isImport
(/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/plugin/modules.js:79:44) at
Object.findCompletions (/Users/jianhuaren/.atom/packages/atom-
ternjs/node_modules/tern/plugin/modules.js:261:18) at Object.signalReturnFirst
(/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/lib/signal.js:34:27) at
Object.findCompletions [as run] (/Users/jianhuaren/.atom/packages/atom-
ternjs/node_modules/tern/lib/tern.js:668:26) at run (/Users/jianhuaren/.atom/packages/atom-
ternjs/node_modules/tern/lib/tern.js:254:30) at Object.exports.withContext 
(/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/lib/infer.js:862:18) at 
/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/lib/tern.js:262:13 at analyzeAll 
(/Users/jianhuaren/.atom/packages/atom-ternjs/node_modules/tern/lib/tern.js:376:15)
```

when i type `var a ` it will display

agileago:
this issue can be closed

PR #650, opened 2015-09-27
fix init is undefined caught exception

agileago:
when type `var a=` it will caught exception. this change will fix it issue #649


marijnh:
Oh, I didn't see this before looking at the bug report, and already pushed an identical patch as 63818e4e5da2

Issue #331, opened 2014-05-11
Running Tests Produces Errors


Mike-E-angelo:
Hello all,

I apologize if this is not the right forum for these types of questions.  If there is a better place to ask this, please kindly point me in the right direction and I will do that from now on.

I am attempting to run Tern's tests from my local install directory, and I am getting errors.  I run "npm install" and then from the root directory of Tern, I execute "node bin/test".  The tests appear to run, but I get exceptions.  Here are a few of them:

angular, line 9: Expression has type
  ?
instead of expected type
  number
angular, line 10: Expression has type
  ?
instead of expected type
  string
angular, line 15: Expression has type
  ?
instead of expected type
  string
angular, line 17: Expression has type
  ?
instead of expected type
  {myColor}
angular, line 18: Expression has type
  ?
instead of expected type
  {myColor}
angular, line 20: Expression has type
  ?
instead of expected type
  string
angular, line 21: Expression has type
  ?
instead of expected type
  string
angular, line 23: Found no docstring instead of expected docstring
  doc for User
angular, line 26: Expression has type
  ?
instead of expected type
  fn(callback: fn()) -> Promise
angular_service.js, line 13: Expression has type
  ?
instead of expected type
  Cart
angular_service.js, line 14: Expression has type
  ?
instead of expected type
  string
browser.js, line 3: Expression has type
  ?
instead of expected type
  Element
browser.js, line 5: Expression has type
  ?
instead of expected type
  Element
browser.js, line 7: Expression has type
  ?
instead of expected type
  string
browser.js, line 11: Expression has type
  ?
instead of expected type
  number
component, line 3: Expression has type
  ?
instead of expected type
  fn() -> string
component, line 4: Expression has type
  ?
instead of expected type
  fn() -> number
def_type_string.js, line 3: Expression has type
  ?
instead of expected type
  string
hint_browser.js, line 2: Completion set failed at hint: document
     got:
  wanted: document, close, closed
hint_browser.js, line 3: Completion set failed at hint: appendChild
     got:
  wanted: appendChild, close, createAttribute
hint_browser.js, line 4: Completion set failed at hint: addEventListener
     got:
  wanted: addEventListener, appendChild
hint_browser.js, line 6: Completion set failed at hint: border
     got:
  wanted: border, borderColor
hint_browser.js, line 9: Completion set failed at hint: undefined
     got:
  wanted: toExponential, toString, toFixed
jquery_event.js, line 5: Expression has type
  ?
instead of expected type
[...]

PR #660, opened 2015-10-01
Fix failing tests on Windows

i-p:
Tests on my Windows 7 machine failed with errors similar to #331.

I've found two issues:

1. my repository was checked out with CRLF line endings, but test runners expect files with LF.
I think the easiest solution is to add .gitattributes file, so that test files are always checked out with LF line endings.

2. typo in `normalizeFilename()`

After these changes all tests are passing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+test/**/* text eol=lf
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks! Merged.

PR #661, opened 2015-10-05
Added new node.js `util` functions

dani-h:
Added definitions to the node `util` module.

- isBoolean
- isNull
- isString
- isNullOrUndefined
- isNumber
- isString
- isSymbol
- isUndefined
- isObject
- isFunction
- isPrimitive
- isBuffer


marijnh:
Thanks! Merged as c061557

PR #664, opened 2015-10-15
[node] data type for "fs" can be String or Buffer

angelozerr:


marijnh:
Thanks. Merged as cc118e2

PR #665, opened 2015-10-15
[browser] Node method like Node#appendChild must use Node instead of …

angelozerr:
…Element


marijnh:
Looks good. Merged as 78347c3

PR #678, opened 2015-11-08
Support @class and @constructor jsdoc attributes

Rantanen:
http://usejsdoc.org/tags-class.html

`@class` and `@constructor` behave the same as `@this`, but unlike `@this`, which needs an explicit type name, these two attributes define the 'this' parameter as the current function without having to implicitly name it.

Only the default syntax without explicit type/name is supported.

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -371,8 +371,19 @@
 
     for (var i = 0; i < comments.length; ++i) {
       var comment = comments[i];
-      var decl = /(?:\n|$|\*)\s*@(type|param|arg(?:ument)?|returns?|this)\s+(.*)/g, m;
+      var decl = /(?:\n|$|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)\s+(.*)/g, m;
       while (m = decl.exec(comment)) {
+        if (m[1] == "class" || m[1] == "constructor") {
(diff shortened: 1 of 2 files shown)


marijnh:
Should `this` really refer to the constructor function in this case? The way I understand it, it should be pointing at an instance of that constructor (i.e. an object derived from its `prototype` property).

Rantanen:
Doesn't the `parseType` take care of that?

```
/** @param {Foo} f */
var method = function(f) {
  // f is instance of 'Foo'
}

/** @class */
var Foo = function() {
  // 'this' is instance of 'Foo'
}
```

.. or then I'm completely failing at jsdoc. Will look whether I'll figure out how the jsdoc type definitions work with self defined types.

marijnh:
I'm going by your tests, which assert the type of this to be `fn()`, which looks suspicious.

Rantanen:
Oh right. That one.

I admit that confused me as well, but figured that was some acorn-specific detail that I didn't understand.

Did more testing. Apparently somewhere there is code that compares the case of the function. 'abc' -> fn(), 'Abc' -> Abc. Fixed tests.

marijnh:
Ah, I see what was going on now -- because the constructors weren't capitalized, the parseType didn't treat them as constructors. I've pushed a somewhat robust implementation in d45e911, and merged your tests in 1ac35c058b2

PR #681, opened 2015-11-09
New deleteDefsByName API for Tern.Server

albertinad:
Hi @marijnh, this PR is a proposal for a new API to Tern.Server. 
I'm using tern to provide hints, and I have an scenario where I dynamically generate tern definitions for specific APIs, and I use `Tern.Server.addDefs` API to update the current Tern.Server instance definitions. Sometimes, those APIs are removed in my application, and I need to tell tern server instance that those definitions should not be used anymore. To do that, today, I have a couple of alternatives:

1. Keep in cache the current configuration for the tern server, create a new instance of `Tern.Server` and pass the cached configuration.
2. Use the implementation proposed in this PR in my implementation that consumes the `Tern.Server` instance APIs.

Since this implementation for removing added definitions could be useful for other applications that uses Tern, I proposed in this PR to extend `Tern.Server` APIs by adding this one.

I have a couple of alternatives on how to implement it. But this one doesn't require any extra change in how defs are handled by tern and def modules.

Thanks!

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -843,6 +843,17 @@ Otherwise, they are added at the back.
 </p>
 </dd>
 <dt>
+<a id="tern.Server.deleteDefsByName"></a> <code>deleteDefsByName(name: string)</code>
+</dt>
+<dd>
+<p>
+Delete
+a set of <a href="#typedef">type definitions</a> from the server, by providing the name, taken from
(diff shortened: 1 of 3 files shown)


marijnh:
Thanks. I guess that makes sense to have. I've merged your work as fb5f0af5ced, and added patch 201fa11 which renames the method to `deleteDefs` and simplifies its implementation a little.

PR #682, opened 2015-11-10
Store the reference httpServer in tern server instance

angelozerr:
This PR gives the capability to retrieve by a tern plugin the httpServer from the server server instance. See discussion at https://github.com/ternjs/tern/issues/679

In my case I need that to use tern with WebSocket to send to the client when Angular modules changes. I'm developping a generic project which provides a new method to the tern server sendToClient in https://github.com/angelozerr/tern-push

### DIFF
--- a/bin/tern
+++ b/bin/tern
@@ -208,6 +208,7 @@ var httpServer = require("http").createServer(function(req, resp) {
     else respondSimple(resp, 400, "Missing query document");
   }
 });
+server.httpServer = httpServer;
 httpServer.listen(port, host, function() {
   var port = httpServer.address().port;
   if (!noPortFile) {


marijnh:
Thanks. Merged as c338edc

angelozerr:
Many thanks @marijnh!

angelozerr:
@marijnh I need to initialize teh WebSocketServer when httpServer is started (for the moment I use the event "beforeLoad" to do that which is not very elegant.

I tell me if you could agree to fire an event with signal before https://github.com/ternjs/tern/blob/master/bin/tern#L226 like 

```javascript
server.signal("httpServerStarted", httpServer)
```

And remove the store of the reference of httpServer. Are you OK with that?

marijnh:
So, ehh, why didn't you wait with sending a patch until you actually had something that you were happy with? I was wondering about exactly that, but figured you didn't need the property on plugin init.

How about adding an option `parent` to the server, where the component that initializes it can store a reference to itself, just like we have on infer contexts? So the option's value would be stored in the server's `parent` property, and the node-based server could store something like `{server: httpServer}` in it. Does that sound like it would help?

angelozerr:
> So, ehh, why didn't you wait with sending a patch until you actually had something that you were happy with? I was wondering about exactly that, but figured you didn't need the property on plugin init.

Sorry @marijnh -( I'm too impatient. My PR works but it'not very clean. Next time I will finish my implementation when I will create PR.

> Does that sound like it would help?

I don't understand very well your proposition. My need is to create a WebSocketServer just after the call of httpServer.listen

angelozerr:
@marijnh I have created a PR https://github.com/ternjs/tern/pull/683 with your idea an dit works great in my case. Hope you will accept it and I'm really sorry with my first PR -(

PR #683, opened 2015-11-10
Set httpServer as parent for tern server	

angelozerr:
This PR implement the idea of @marijnh https://github.com/ternjs/tern/pull/682#issuecomment-155397016.

When tern server is started with node.js, the tern server has parent with httpServer. It means that a tern plugin can get the httpServer like this : 

```javascript
tern.registerPlugin("push", function(server, options) {
  var httpServer = server.parent;
});
```

### DIFF
--- a/bin/tern
+++ b/bin/tern
@@ -144,9 +144,28 @@ if (projectDir) {
   projectDir = process.cwd();
   var config = defaultConfig;
 }
-var server = startServer(projectDir, config);
 
-function startServer(dir, config) {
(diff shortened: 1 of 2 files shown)


angelozerr:
@marijnh is there any chance that you review my PR please. I would like to support WebSocket inside tern.java. Thanks!

marijnh:
Merged as 00ad46daa0b0, updated in db3f169390c to make it more future-proof.

angelozerr:
Many thanks @marijnh!

PR #687, opened 2015-11-20
Fix lexical binding for Emacs plugin

### LINKED ISSUE #685: lexcial-binding stripped from tern.el file-header
when using:

```cl
(package-install-file "/path/to/tern.el")
```

the installed package is missing `;;; -*- lexical-binding: t -*-` which causes it to malfunction
Emacs probably removes the line because:

```
;;; filename --- description

The description should be contained in one line. If the file needs a ‘-*-’ specification, put it after description. If this would make the first line too long, use a Local Variables section at the end of the file. 
```

see http://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

could you correct that?

vyp:
See [1].

Closes #685.

[1]: http://www.gnu.org/software/emacs/manual/html_node/elisp/Library-Headers.html

---

cc @steckerhalter, does this solve the problem?

### DIFF
--- a/emacs/tern-auto-complete.el
+++ b/emacs/tern-auto-complete.el
@@ -1,5 +1,4 @@
-;;; -*- lexical-binding: t -*-
-;;; tern-auto-complete.el --- Tern Completion by auto-complete.el
+;;; tern-auto-complete.el --- Tern Completion by auto-complete.el -*- lexical-binding: t -*-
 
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks! Merged as 1e9586a

vyp:
thanks!

Issue #690, opened 2015-11-28
Absolute paths in 'files' section of JSON query


puremourning:
Hi there,

TL;DR - Adding tern support to YouCompleteMe, need to use absolute paths for the 'files' part of the query. Should this work? Can it?

First a bit of background: I'm a maintainer of [YouCompleteMe](http://github.com/Valloric/YouCompleteMe), and I'm currently working on adding javascript semantic support to [ycmd](http://gitbub.com/Valloric/ycmd), the multi-language code comprehension server that powers it. Tern is very cool and we're really excited about integrating it thanks to your super simple and well-documented, REST API. I got the basics including code completion, docs, goto, etc. working in just a couple of hours.

However, in testing I've noticed a snag related to dirty buffers. When issuing requests, I'm specifying absolute paths in the 'files' object, as ycmd has no knowledge about the project structure. I expected that the server would be able to resolve the absolute path to the relative paths that it uses internally (if that is indeed the case).

This appears to be resolved by [tern_for_vim](http://github.com/ternjs/tern_for_vim) by making all paths relative to the (assumed) project directory, but as a user of the API, I don't really want to have to know anything about tern's project structure.

To be a specific, consider the following filesystem structure, rooted under `/Users/ben/Development/tmp/js`:

```
/Users/ben/Development/tmp/js/lib/a_module.js
/Users/ben/Development/tmp/js/test_file.js
/Users/ben/Development/tmp/js/.tern-project
```

With these contents:

- `./lib/a_module.js`:

```javascript
define( [], function() {
    return {
        /**
         * This function takes a number and does some math on it
         */
        do_math: function( how_much ) {
            return how_much * 0.001;
        }
    }
} )
```

- `./test_file.js`

```javascript
define( [ 'lib/a_module.js' ], function( a_module ) {
  a_module.do_math( 10 );
} );
```

- `./.tern-project`

```json
{
  "libs": [
  ],
  "plugins": {
    "requirejs": {
      "baseURL": "./",
      "paths": {}
    }
  }
}
```

So the idea is pretty simple - a requirejs module which exports an object with a single function in it.

If I make a change to the `lib/a_module.js` and don't save it, ycmd can send a request containing the dirty buffer. This is great, but I'm seeing a problem when I do this, the problem is that the updated (dirty) code is not used. For example if I change it like this:

```javascript
define( [], function() {
    return {
        /**
         * This function takes a number and does some math on it.
         * And some more text.
         */
        do_math: function( how_much ) {
[...]

PR #695, opened 2015-12-06
[RFC] Allow absolute paths to be used in query

puremourning:
This relates to #690 where a more complete use-case example can be seen.

We're trying to integrate tern.js with YouCompleteMe, a Vim completion plugin used by tens of thousands of users.

The primary use-case for this is that users of the API don't (or, shouldn't)
need to know anything about `.tern-project` and tern's internal project
structure. The simplest thing for editors to do is to just supply the abosolute
path to dirty buffers (in the files section of the query) and have tern
determine if/that they are also included by things like require.js.

A typical example might be a file containing `require('lib/test.js', ...)` and
an API user supplying an unsaved version of 'lib/test.js' with its absolute path
as the filename.

Submitting this in the hope of getting some feedback on the approach.

Things TODO:
* [ ] decide if always storing relative paths is correct, rather than always storing absolute paths (which seems simpler, and less brittle)
* [ ] find a way of testing this automatically (I've started [some work](https://github.com/puremourning/tern/commit/1e190ec4622d37752253fa614f257246988ae8a5) on documenting the test suite and learning how it fits together)
* [ ] anything else you might think is required.

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -129,7 +129,7 @@
     addFile: function(name, /*optional*/ text, parent) {
       // Don't crash when sloppy plugins pass non-existent parent ids
       if (parent && !(parent in this.fileMap)) parent = null;
-      if (!(name in this.fileMap))
+      if (!(this.normalizeFilename(name) in this.fileMap))
         name = this.normalizeFilename(name)
       ensureFile(this, name, parent, text);
     },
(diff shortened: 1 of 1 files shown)


marijnh:
Superseded by patch 24044a4781089b3

PR #697, opened 2015-12-09
Fix flattenPath to correctly process parent (..) dir.

happibum:
Currently, flattenPath('/a/b/c/../../d/e/f') returns 'a/b/../d/e/f'.
This patch makes it return 'a/d/e/f'.


marijnh:
Thanks! Merged as 6bf22c6

PR #698, opened 2015-12-11
Fix default configuration detection on Windows.

CentricStorm:
The configuration file was never found because HOMEPATH doesn't return a full path on Windows (missing drive letter).

HOME works for both OS X and Linux, USERPROFILE works on Windows.


marijnh:
Thanks! Merged as f585fe7

PR #703, opened 2015-12-23
Improve find references query.

piotrtomiak:
- support class method definitions
- take type hierarchy into account when looking for references
- allow for limited search within a single file ("onlySourceFile" query parameter)
- update tests

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -1892,6 +1892,11 @@
       if (propN == "<i>") return ANull;
       return findByPropertyName(propN);
     },
+    MethodDefinition: function(node, scope) {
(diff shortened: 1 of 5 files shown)


piotrtomiak:
Corrected find type method for MethodDeclaration to work with findDef and doc properly (force pushed). Added scope parameter to callback function in findPropRefs, which is exported and is not equivalent to findRefs in that matter (I need that scope for my custom queries).

Let me know if you would wish this contribution to be split into several smaller ones.

marijnh:
Thanks! Looks good. Merged as a9ba18e049

piotrtomiak:
Thanks!

PR #704, opened 2015-12-28
Fixing what I think is a typo

tkrugg:


marijnh:
Thanks! The intended word was 'in', which I've added in 2bd7cbc

PR #728, opened 2016-03-13
Fix missing = in docs

treythomas123:


marijnh:
Thanks! Merged as 36c2ead

PR #738, opened 2016-03-31
Add additional logging on process exit reason

vincentwoo:
re: my comment https://github.com/ternjs/tern_for_vim/issues/16#issuecomment-203734505, I think this would have greatly improved my ability to figure out what was going wrong with trying to run tern.


marijnh:
Seems like a good idea. Merged as 1372f7f

PR #741, opened 2016-04-03
added info and link to TextMate plugin

fab1an:


marijnh:
Thanks. Merged as 91f7692

PR #742, opened 2016-04-04
added --ignore-stdin flag, fixes #740

### LINKED ISSUE #740: Run tern without stdin
I'm trying to write the TextMate plugin for ternjs.
I'm looking for the `.tern-port` file and start the server if it's not found.

My problem is that tern always exits as soon as the script exits because it's stdin pipe breaks. Is there a way around this?

fab1an:


marijnh:
Thank you. Merged as 0527ed226d

PR #743, opened 2016-04-05
Resolve plugin from the current working directory

oligot:
While trying to test a tern plugin I'm writing, I encountered this error
`Failed to find plugin ...`
By using the [resolve-cwd](https://github.com/sindresorhus/resolve-cwd) module, this now works as expected, meaning I
can include my tern plugin in a given project, launch tern from the
project root directory, and make it work :smirk:


marijnh:
Good idea. Merged as df71c215b9a1, and updated a bit in d3bdf8a1377

PR #746, opened 2016-04-08
Export refFindWalker

jacksonrayhamilton:
I am currently developing a plugin that needs to modify this visitor in order to define behavior for a syntax extension to ECMAScript. Will you please expose this object so I may do that?

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -2075,7 +2075,7 @@
     type.gatherProperties(f, 0);
   };
 
-  var refFindWalker = walk.make({}, searchVisitor);
+  var refFindWalker = exports.refFindWalker = walk.make({}, searchVisitor);
 
   exports.findRefs = function(ast, baseScope, name, refScope, f) {
     refFindWalker.Identifier = refFindWalker.VariablePattern = function(node, scope) {


marijnh:
That object is only made in order to override two of its fields inside of `findRefs`. I really can't think of a single sensible use of it from an external module. `searchVisitor` is already exported -- doesn't it make more sense to build your own walker based on that?

jacksonrayhamilton:
Here is where I would like to use it in my plugin: https://github.com/jacksonrayhamilton/tern-jsx/blob/9ab65719c6bb5ea2c66f3fc35373203a440a5c78/jsx.js#L50

Without patching this walker, attempts to rename variables fail if the file uses extended syntax. e.g. If I tried to rename `Element` below,

```jsx
var Element;
var App = <Element/>;
```

then I would get this error,

```
TypeError: visitor[(override || node.type)] is not a function
```

 Is there a better way to deal with that issue than modifying `refFindWalker`?

marijnh:
It seems like you'd get a lot further by patching the original `base` walker functions in acorn, before Tern derives its walkers from them. Needing access to every walker object is going to break every time a plugin uses a walker, isn't it?

jacksonrayhamilton:
> It seems like you'd get a lot further by patching the original base walker functions in acorn, before Tern derives its walkers from them.

Yeah, that sounds better. Is there a way to do that?

Alternatively, if `walk.make` used `Object.create` to inherit from `walk.base`, then modifications to `walk.base` could propagate to walkers after their creation.

> Needing access to every walker object is going to break every time a plugin uses a walker, isn't it?

Yep, probably will break.

marijnh:
> Yeah, that sounds better. Is there a way to do that?

If you have control over the way Tern is loaded, sure, just load Acorn's walker first and add the new walker methods to its base walker.

jacksonrayhamilton:
Are you suggesting I add code to my copy of `bin/tern` to patch the base walker? I'm hoping for a solution that will allow this plugin to work without users specially modifying Tern.

angelozerr:
Hi @jacksonrayhamilton it seems that we have similar project, see https://github.com/angelozerr/tern-react/blob/master/react.js but I have stopped this project since acorn-jsx doesn't support acorn loose (see https://github.com/RReverser/acorn-jsx/issues/32) Have you fixed this problem?

jacksonrayhamilton:
Hi @angelozerr, I was planning on contacting you regarding the relationship between our plugins. I think it would be off-topic to discuss that here.

angelozerr:
> I think it would be off-topic to discuss that here.

yes sure! Thanks for your work.

marijnh:
> Are you suggesting I add code to my copy of bin/tern to patch the base walker?

No, I'm suggesting you load your file, which adds stuff to the base walker, before you load Tern.

jacksonrayhamilton:
The first thing `bin/tern` does is `require("../lib/tern")`, and that file does `require("./infer")`, and that file starts creating walkers. So as far as I can tell, to mutate the walker before Tern loads, I'd have to create a wrapper script like this, and put it on my PATH as "tern" so my editor can find it:

```js
#!/usr/bin/env node
var walk = require('tern/node_modules/acorn/dist/walk');
var inject = require('tern-jsx/inject');
inject(walk.base);
require('tern/bin/tern');
```

That would not be very ideal... this script would compete with the globally-installed `tern`, or I would have to ask users to install a wrapper around the `tern` package instead of the actual `tern`.

Did you mean something else? If so, will you please explain in more detail?

marijnh:
That's why I said 'if you have control over the way tern is loaded before'.

If you can get something working by making `walk.make` use `Object.create`, I'm also okay with that -- you can submit a PR to acorn to make that change if it helps.

jacksonrayhamilton:
> That's why I said 'if you have control over the way tern is loaded before'.

Ah, sorry, I did not quite take that in.

> If you can get something working by making `walk.make` use `Object.create`, I'm also okay with that

Great, I'll see what I can come up with.

However, given the way `refFindWalker` is currently used, I will probably still want to modify it like I do here: https://github.com/jacksonrayhamilton/tern-jsx/blob/master/jsx.js#L54 And as a heads-up, the same may go for a few other internal structures (as I'm [currently](https://github.com/jacksonrayhamilton/tern-jsx/blob/4d5d7d4c318941808f5e6b09feac29fc906d7c72/jsx-test.js#L186) [discovering](https://github.com/jacksonrayhamilton/tern-jsx/blob/4d5d7d4c318941808f5e6b09feac29fc906d7c72/jsx-test.js#L209) in my test suite), despite the improvement that may be accomplished by updating `walk.make`.

marijnh:
> However, given the way refFindWalker is currently used, I will probably still want to modify it like I do here:

Seems like it might make sense to make `JSXIdentifier` call through to `Identifier` when not overridden.

jacksonrayhamilton:
> Seems like it might make sense to make `JSXIdentifier` call through to `Identifier` when not overridden.

True, if I were modifying `walk.base` and it were a prototype, then in my situation I could probably do:

```js
walk.base.JSXIdentifier = function () {
  return this.Identifier.apply(this, arguments);
};
```

However, if the handler were detached and the context unspecified, it'd break (due to `this` semantics). Calling through in this manner also seems like it may be trying to serve a particular use case in Tern where the handler is defined ad-hoc. As I derive from the conventions of [walk/index.js](https://github.com/ternjs/acorn/blob/master/src/walk/index.js), I ought to assign an `ignore` function as the default value of the `JSXIdentifier` handler.

jacksonrayhamilton:
Indeed, if we make `walk.base` a source of retroactive heredity then it seems like I can use the already-exposed `searchVisitor` for this:

```js
infer.searchVisitor.JSXIdentifier = function () {
  return this.Identifier.apply(this, arguments);
};
```

Some of the methods on `walk` do detach the handler, but thankfully that isn't true of the methods currently receiving `infer.searchVisitor`. I'd feel more comfortable if code like this:

```js
let found = visitors[type]
if (found) found(node, st)
```

Were more defensively written like this:

```js
let found = visitors[type]
if (found) found.call(visitors, node, st)
```

marijnh:
I actually meant something more along the lines of

```
walk.base.JSXIdentifier = function(node, st, c) {
  c(node, st, "Identifier")
}
```

That doesn't rely on `this` being the right object, and is more like the way other 'forwarding' works in the walker.

jacksonrayhamilton:
Right, but here's why I don't think that can work:

Yes, JSXIdentifier implements the same interface as Identifier. JSXIdentifier and Identifier often represent variables in scope, so in some cases (like searching for a definition or reference) then they are treated the same. But it is not always true that JSXIdentifier should be defined in terms of Identifier.

For instance, in another plugin I am working on, [tern-context-coloring](https://github.com/jacksonrayhamilton/tern-context-coloring), I intend to color Identifier nodes by their scope. White is global scope, blue is level 1, green is level 2, and so. Modifying `walk.base` to call through will result in also coloring JSXIdentifier nodes by scope. And indeed, sometimes a JSXIdentifier references a variable in scope. So, I observe the following (unintended) behavior in my plugin (i.e., "Form," a JSXIdentifier, is colored blue in a JSXElement):

<p align="center">unexpected yet correct scope coloring</p>

But other times, JSXIdentifier refers to an HTML element. Is that a variable in scope? Nope. But my plugin assumes that these JSXIdentifier refer to global variables ("div" is colored white, which means it is global):

<p align="center">accidental global variable coloring</p>

So what I would have preferred in this case were that JSXIdentifier just be ignored, so I could get this "neutral" coloring:

<p align="center">no special JSX coloring</p>

If I wanted to prevent that unintentional coloring, I could apply the knowledge I have and make a special redefinition of JSXIdentifier in tern-context-coloring. But it doesn't end there; my non-standard syntax extension could impose special handling requirements on _any_ Tern plugin in a similar manner.

I guess what this boils down to is that JSXIdentifier =/= Identifier, so it's inappropriate to call through to it by default in `walk.base`. And I'm not arguing with any particular goal in mind right now, except to try and explain my reasoning in case this mattered sometime in the future.

marijnh:
Having the default JSXIdentifier defer to Identifier doesn't mean you can't override it. If want specific behavior, as you seem to want in your example, just override it (JSXIdentifier) with an implementation of that behavior.

jacksonrayhamilton:
I forgot to deliver an important part of my argument: If I define `walk.base.JSXIdentifier` as `ignore`, then `JSXIdentifier` is not automatically assumed to be `Identifier`, which results in other plugins (like tern-context-coloring), which don't necessarily know about JSX, to just skip over the syntax (as in the third screenshot). I think that is best.

As I attempt to illustrate in the example, not all plugins are _prepared_ for this node type. If they know about it, sure, they can define a walker specifically for it that does something additive. But I believe, that by default, it is better to presume as little as possible about the contexts JSXIdentifier will be used; this way, authors can choose to _add_ support for non-standard syntax, rather than receive bug reports forcing them to explicitly _ignore_ my non-standard syntax, and Joe's and Sue's non-standard syntaxes, because they break something.

marijnh:
I guess this is kind of an awkward case in that it _sometimes_ behaves just like an identifier, but not always, so indeed, it might cause problems to treat it as one. So sure, if making the default behavior be to ignore the node works, that's probably better.

jacksonrayhamilton:
Okay, point made. :P Also, the `walk.make` update has been working for me on my Tern fork, so I'll consider this issue solved.

PR #750, opened 2016-04-11
Adding vy editor to the list of editors using tern.

iogf:


marijnh:
Well, this reads more like an advertisement for vy than any kind of useful information. If the plugin is built-in, people don't need to read about it in the manual. I've added a link to the README, I think that's enough in this case.

iogf:
It seems the editor brackets has a built-in plugin and it is listed in the page.

http://ternjs.net/

As vy tern plugin is very well designed and it is about 100 lines of code compared to other editor implementations i thought it is due to tern having a good interface and having vy tern plugin listed there it would be something for you guys to be proud of.

PR #754, opened 2016-04-19
Catch scope

jacksonrayhamilton:
These changes make it so the scope of the `catch` parameter appears in the scope hierarchy. Previously, the `catch` parameter was treated as if it were declared with `var`. Therefore, if there were multiple `catch` clauses with the same parameter name in the same scope, then attempts to find the definition of the second would lead back to the first.

One motivation for this change is to be able to properly locate the definition of the parameter. Another is so that I can write a plugin which does scope analysis and `catch` parameters can be handled correctly.

In order to walk `CatchClause`, I had to update to the latest Acorn. I am not sure if the major version bump is cause for concern, but maybe it is if your dependents are requiring your copy of `acorn`. If that worries you, maybe we can backport the `CatchClause` support to `v2.x`, or see if this could still be accomplished with `TryStatement`.

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -924,11 +924,12 @@
 
   // SCOPES
 
-  var Scope = exports.Scope = function(prev, originNode, isBlock) {
+  var Scope = exports.Scope = function(prev, originNode, isBlock, isCatch) {
     Obj.call(this, prev || true);
     this.prev = prev;
     this.originNode = originNode
     this.isBlock = !!isBlock
+    this.isCatch = !!isCatch
   };
   Scope.prototype = extend(Obj.prototype, {
     constructor: Scope,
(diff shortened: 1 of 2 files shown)


jacksonrayhamilton:
Also, if you want, I can add some test cases for this. I wanted to, but wasn't quite sure where they'd go or how they'd be written in this project's test runner.

marijnh:
There (unfortunately) aren't any tests for the walker. If you want to start adding some, that'd definitely be welcome (the test runner is a bit ad-hoc, but it should be possible to add arbitrary tests with it)

marijnh:
I've merged this as d2414a66d2d300...b4bd2fcd9fc6 and added a (somewhat trivial) test in 10744d53f082346a

PR #766, opened 2016-05-09
Fix for missing "this." in delFile loop

vincentwoo:
see https://github.com/ternjs/tern/issues/764


marijnh:
Thanks. I fired off 3da8392 before I noticed this PR.

PR #776, opened 2016-05-17
node plugin: updating links

### LINKED ISSUE #775: node plugin: outdated defs
There are some defs that are outdated, for example:

- `console` module uses wrong URL (https://nodejs.org/api/stdio.html#....) while current one look like (https://nodejs.org/api/console.html#....)
- `events` module uses wrong hashes, for example https://nodejs.org/api/events.html#events_class_events_eventemitter while expected https://nodejs.org/api/events.html#events_class_eventemitter

alexsaveliev:
- fixes #775

### DIFF
--- a/plugin/node.js
+++ b/plugin/node.js
@@ -27,27 +27,27 @@
           prototype: {
             addListener: {
               "!type": "fn(event: string, listener: fn())",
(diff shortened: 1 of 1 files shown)


marijnh:
Appreciated! Merged as f566b5b

Issue #777, opened 2016-05-19
Maximum call stack size exceeded when querying definition of 'foo' in {foo:bar}


alexsaveliev:
#### How to repeat
- create `data.js` with the following content
```
function main() {
	return function next() {
		return {foo: 1};
	};
}
```
- try to query definition of `foo`
I'm using the following code:
```
var fs = require('fs');
var tern = require('tern');
var walk = require('acorn/dist/walk');

var server = new tern.Server({
	debug: true,
	getFile: function(file) {
    	return fs.readFileSync(file, "utf8");
    }
});

server.addFile("data.js");

var visitor = {
    ObjectExpression: function(node, kind, c) {
        for (var i = 0; i < node.properties.length; ++i) {
            c(node.properties[i].key, kind);
        }
    },
    Identifier: function(node, kind, c) {
    	var request = { query: {type: 'definition',
      		start: node.start,
      		end: node.end,
      		file: node.sourceFile.name
    	}};
    	server.request(request, function(err, data) {
    		if (err) {
    			console.log(err);
    		}
    	});
    }
};

server.flush(function() {
	server.files.forEach(function(file) {
		walk.recursive(file.ast, null, visitor);
	});	
});
```

#### Expected result

There is no definition of `foo` available so server should return no results

#### Actual result
```
RangeError: Maximum call stack size exceeded
    at c (D:\exp\trav\node_modules\acorn\dist\walk.js:113:17)
    at Object.base.Function (D:\exp\trav\node_modules\acorn\dist\walk.js:260:16)
    at c (D:\exp\trav\node_modules\acorn\dist\walk.js:116:17)
    at Object.base.FunctionDeclaration (D:\exp\trav\node_modules\acorn\dist\walk.js:247:10)
    at c (D:\exp\trav\node_modules\acorn\dist\walk.js:116:17)
    at Object.skipThrough (D:\exp\trav\node_modules\acorn\dist\walk.js:169:3)
    at c (D:\exp\trav\node_modules\acorn\dist\walk.js:116:17)
    at Object.base.ReturnStatement.base.YieldExpression (D:\exp\trav\node_modules\acorn\dist\walk.js:212:22)
    at c (D:\exp\trav\node_modules\acorn\dist\walk.js:116:17)
    at Object.skipThrough (D:\exp\trav\node_modules\acorn\dist\walk.js:169:3)
{ [RangeError: Maximum call stack size exceeded] undefined: true }
```

marijnh:
Fixed via #778

PR #778, opened 2016-05-19
Fixes endless recursion when evaluating type

alexsaveliev:
Fixes endless recursion when trying to evaluate object's key type when object is enclosed into two functions (#777).
Test case
```
function main() {
	return function next() {
		return {foo: 1};
	};
}
```
Problem happened because once we reached "next" function expression, `findNodeAround(.., "function")` returned the same expression causing infinite loop

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -2040,7 +2040,10 @@
       }
     },
     ReturnStatement: function(_parent, node, get) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Merged as 49c2680

Issue #780, opened 2016-05-19
All proposals shown inside string literal


mrennie:
Consider the following snippet:

`var foo = "";`

if you activate content assist inside the string literal you get all of the JS proposals when the complete_strings plug-in is not in use. I would expect to get no proposals.

marijnh:
See my answer to #757

PR #781, opened 2016-05-19
All proposals shown inside string literal #780

mrennie:
https://github.com/ternjs/tern/issues/780

If there are no loaded plugins that handle completions inside literals, we should return none.

Related Orion bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=492470

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -713,6 +713,14 @@
     // expression or an object literal.
     if (exprAt) {
       var exprNode = exprAt.node;
+      if(exprNode.type === 'Literal' && typeof exprNode.value === 'string') {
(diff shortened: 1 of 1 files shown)


marijnh:
This is something that I think you should solve on the client side. Or, if that's difficult, add this as a query option that defaults to false. I want the default behavior to stay what it is now.

mrennie:
Ok, I'll update the pull request to add a query flag. Any particular name you would prefer for the new flag?

marijnh:
`inLiterals: false` sounds like a descriptive name (where anything non-false counts as true).

mrennie:
I'm going to close this pull request and open a new one - to avoid the merge commit.

PR #785, opened 2016-06-01
Add missing keywords when ecmaVersion = 6

### LINKED ISSUE #784: Missing keyword autocompletion for ES6
I noticed that `let` `const` and `import/export` can't be auto-completed

    var jsKeywords = ("break do instanceof typeof case else new var " +  
        "catch finally return void continue for switch while debugger " +  
        "function this with default if throw delete in try").split(" ");

seems like they are missing in declaration.

tommytroylin:
Add missing keyword autocompletion for ES6, e.g. let const static etc.

Only works when ecmaVersion = 6
nothing has changed in ES5

Fix #784

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -653,6 +653,10 @@
   var jsKeywords = ("break do instanceof typeof case else new var " +
     "catch finally return void continue for switch while debugger " +
     "function this with default if throw delete in try").split(" ");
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as 0de94999c, updated in ead32cf20d97

Issue #769, opened 2016-05-15
Doc comments not picked up with ES2015 export statement


akim-mcmath:
Tern doesn't seem to pick up the JSDoc comment in the following case:

```js
/**
 * Divides one number by another.
 *
 * @param   {number}  x  Numerator.
 * @param   {number}  y  Denominator.
 * @return  {number}     Quotient.
 */
export function divide(x, y) {
  return x / y;
}
```

It behaves just as if the comment wasn't there. It recognizes the `divide()` function as an export,
but it doesn't get the description and type information from the comment.

Oddly, however, the following works fine:

```js
/**
 * Divides one number by another.
 *
 * @param   {number}  x  Numerator.
 * @param   {number}  y  Denominator.
 * @return  {number}     Quotient.
 */
function divide(x, y) {
  return x / y;
}

export {divide};
```

Is this a bug, or am I missing something?

marijnh:
Fixed in 453f6fd

akim-mcmath:
@marijnh Thank you very much for this, and for your work on this fantastic project.

alisd23:
Having the same issue with v0.21.0, except with variable declarations:

The following does not recognise the typing of `someString`
```js
/** @type {string} */
export let someString;
```

Yet this does:
```js
/** @type {string} */
let someString;
export someString;
```

PR #789, opened 2016-06-06
Doc comments not picked up with ES2015 export statement #769

mrennie:
https://github.com/ternjs/tern/issues/769

Adds support in the doc_comment plugin to attach comments to, and process the comments on ExportNamedDeclaration and ExportDefaultDeclaration nodes.


marijnh:
Thanks for looking into that! merged as 453f6fd

PR #793, opened 2016-06-24
Added null check in the registered "getSymbol" function

alexsaveliev:
Sometimes `argNodes` may be `null` causing
```
"TypeError: Cannot read property 'length' of null
    at .../node_modules/tern/lib/def.js:668:17"
```
Added check to prevent this error


marijnh:
Thanks, merged (finally) as e878e67

PR #796, opened 2016-06-29
All Proposals shown inside string literal #780

mrennie:
Add new flag that defaults to true (show proposals in literals)

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -715,12 +715,11 @@
     // expression or an object literal.
     if (exprAt) {
       var exprNode = exprAt.node;
-      if(query.inLiteral === false && exprNode.type === 'Literal' && typeof exprNode.value === 'string') {
-        //if one of the plug-ins has not handled in-literal completions, return none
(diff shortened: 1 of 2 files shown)


lib/tern.js
```diff
@@ -715,6 +715,14 @@
     // expression or an object literal.
     if (exprAt) {
       var exprNode = exprAt.node;
+      if(query.inLiteral === false && exprNode.type === 'Literal' && typeof exprNode.value === 'string') {
```
marijnh: Wouldn't it make sense to have the same flag also apply to regexp literals?

lib/tern.js
```diff
@@ -715,6 +715,14 @@
...
+        //if one of the plug-ins has not handled in-literal completions, return none
+        return {
+        	 start: outputPos(query, file, wordStart),
+           end: outputPos(query, file, wordEnd),
+           completions: []
+        };
```
marijnh: (Try to have consistent indentation -- two spaces per block.)

marijnh:
Merged as fb585f6

Issue #797, opened 2016-06-29
Proposals with same names but different origins can be ignored


mrennie:
If you have two plugins that return a similarly named proposal, the one that was inferred or loaded first (from a definition file) is the only one that is returned, see line 795 onward in tern.js:

```for (var prop in srv.cx.props) gather(prop, srv.cx.props[prop][0], 0);```

That loop should walk all of the entries in ```srv.cx.props[prop]``` and gather them all if they come from different origins (and we are asking for origin information).

See Eclipse Orion bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=496955 for a concrete example. I'll open a pull request for how we fixed the problem in Orion.

PR #798, opened 2016-06-29
Proposals with same names but different origins can be ignored #797

mrennie:
Check all properties, not just the first, and check origins before
leaving out same-named properties

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -660,9 +660,14 @@
     var typeInfo = query.types || query.docs || query.urls || query.origins;
     var wrapAsObjs = typeInfo || query.depths;
 
-    for (var i = 0; i < completions.length; ++i) {
+    for (var i = 0, len = completions.length; i < len; i++) {
       var c = completions[i];
-      if ((wrapAsObjs ? c.name : c) == name) return;
+      if ((wrapAsObjs ? c.name : c) == name) {
+        if(query.origins && aval && c.origin !== aval.origin) {
+          continue;
+        }
+        return;
+      }
     }
     var rec = wrapAsObjs ? {name: name} : name;
     completions.push(rec);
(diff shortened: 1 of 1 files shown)


lib/tern.js
```diff
@@ -660,9 +660,14 @@
     var typeInfo = query.types || query.docs || query.urls || query.origins;
     var wrapAsObjs = typeInfo || query.depths;
 
-    for (var i = 0; i < completions.length; ++i) {
+    for (var i = 0, len = completions.length; i < len; i++) {
```
marijnh: This is an [unnecessary](https://www.eternalcoding.com/?p=33) optimization which has nothing to do with the actual change you're making in this patch.

lib/tern.js
```diff
@@ -660,9 +660,14 @@
...
       var c = completions[i];
-      if ((wrapAsObjs ? c.name : c) == name) return;
+      if ((wrapAsObjs ? c.name : c) == name) {
+        if(query.origins && aval && c.origin !== aval.origin) {
```
marijnh: Why not just add the inverse of this condition to the `if` above, rather than introducing extra control flow?

lib/tern.js
```diff
@@ -761,8 +766,13 @@
...
         objType.guessProperties(function(p, o, d) {if (p != prop && p != "✖") gather(p, o, d);});
-      if (!completions.length && word.length >= 2 && query.guess !== false)
-        for (var prop in srv.cx.props) gather(prop, srv.cx.props[prop][0], 0);
+      if (!completions.length && word.length >= 2 && query.guess !== false) {
+        Object.keys(srv.cx.props).forEach(function(prop) {
+          srv.cx.props[prop].forEach(function(type) {
+    	      gather(prop, type, 0);	
+    	    });
+    	  });
+      }
```
marijnh: Is it really a good idea to start adding guessed properties multiple times? That has the potential to create a lot of noise, for very little gain (we don't know the accurate anyway).

PR #800, opened 2016-07-02
Update outdated link for path join method for the node.js plugin

antonina-cherednichenko:


marijnh:
Thank you. Merged as 0832b94

PR #804, opened 2016-07-17
Always determine `tern-project-dir' just-in-time

jacksonrayhamilton:
Fixes https://github.com/ternjs/tern/issues/756.


marijnh:
Well, this is going to make a bunch of syscalls, possibly even touching disk as it walks up the directory tree, every time Tern does anything. I'm not sure I'm fine with that. Isn't this cure as least as bad as the problem? Could we at least cache a value and only recompute it when `buffer-file-name` changes?

jacksonrayhamilton:
Sure, we could track `buffer-file-name` instead.

jacksonrayhamilton:
I am actually a little surprised this works (or at least seems to), because it looks like the same Tern  server instance is used even after the file's project directory name changes, yet the Tern server code only figures out the project directory [once when it starts up](https://github.com/jacksonrayhamilton/tern/blob/0122f3830c52e8e1b89440900850b581539e9791/bin/tern#L146)... This makes me a little weary of suggesting this change.

marijnh:
Thanks! Merged as 95dfbf405...a5a5ba61ae3

marijnh:
Oops, I merged before reading your last comment. A solution to that might be to move the code that fetches the port and the code that determines the directory into a single function, which makes sure to start a new server when the filename changes? Or simply also invalidate `tern-known-port` using `buffer-file-name` as key?

jacksonrayhamilton:
I think there are two cases to consider:

1. A file is moved to another project.
2. A file's project's directory is renamed.

We can easily support the first case by invalidating `tern-known-port` using `tern-project-dir` as key, since `tern-project-dir` will determine if and where `.tern-port` should be sought again.

The second case is trickier. The difference is significant, because the invalidity of Tern's `projectDir` will probably eventually cause bugs. The second case also can't be distinguished by the editor, since all the editor knows is that file path changed, not that the `.tern-port` file now points to the wrong server. We could treat `.tern-port` as truth, blindly connecting to the port it declares, and blame the user for "deliberately misconfiguring his directory." A more forgiving solution would be to store the server `projectDir` in a temporary file, so the editor could determine whether to connect to an existing server or to start another one.

jacksonrayhamilton:
[This commit](https://github.com/jacksonrayhamilton/tern/commit/75bcd9b342a5477c10cf1845132acb721996efbb) ought to satisfy the first case. (Those I'll admit that these state variables are grossing me out.) Up for another PR?

marijnh:
I don't think the second case is worth worrying a lot about. Your patch looks good. I merged it as 7d1c273

PR #805, opened 2016-07-18
support webpack.config.js for webpack plugin

chemzqm:
This PR use latest `enhanced-resolve` module, supports:
* Load `webpack.config.js` from the same folder
* Same default options as webpack itself.
* Configuration like `root` `alias` `packageMains ` etc.
* es6 `import` support by using `es_modules` plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "acorn": "^3.2.0",
-    "enhanced-resolve": "^0.9.1",
+    "enhanced-resolve": "^2.2.2",
     "glob": "3",
     "minimatch": "0.2",
     "resolve-from": "2.0.0"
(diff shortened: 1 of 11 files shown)


marijnh:
Looks good. Could you add a few tests to make sure future changes don't end up breaking this?

karlhorky:
Would it make sense to make the configuration file name configurable?

marijnh:
Thanks! Merged as a5a5ba61ae3a5...b445a26a85e3349

(If the answer to @karlhorky 's question is yes, let's create a different issue or PR for that.)

chemzqm:
@karlhorky added in #807

karlhorky:
Awesome, nice one @chemzqm

PR #807, opened 2016-07-19
add `configPath` option support for webpack plugin

chemzqm:
`configPath` is a file path relative to project directory, eg: `./webpack.dev.js`.
`configPath` default to `./webpack.config.js`


marijnh:
Thank you. Merged as 18932c4

PR #809, opened 2016-07-26
RegExp prototype def update to include overridden 'toString' method

swmitra:
This PR updates 'RegExp' prototype definition to include 'toString' method. 
Reference : <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/toString>

@marijnh Can you please have a look and merge if the change looks fine.

### DIFF
--- a/defs/ecma5.json
+++ b/defs/ecma5.json
@@ -481,6 +481,11 @@
         "!url": "https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/RegExp/test",
         "!doc": "Executes the search for a match between a regular expression and a specified string. Returns true or false."
       },
+      "toString": {
+        "!type": "fn() -> string",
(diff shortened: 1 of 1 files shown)


marijnh:
`toString` is already in `Object.prototype`, so I've avoided adding it to the various prototypes that derive from that, since that doesn't really add any new information.

swmitra:
Thanks a lot @marijnh for review. That was really quick :+1: 
The problem what I have is, when I create a RegExp object and try to provide hints on the instance var upon typing '.' , the hint list doesn't include 'toString'. Ideally, as 'toString' is overridden in 'RegExp' , that kind of make 'toString' an explicit behavior in case of 'RegExp' instead of an inherited behavior. So, the immediate hint list from Tern should include the 'toString' method after inferring the member to be of type 'RegExp' I guess. Do let me know your thoughts on this.

marijnh:
That is indeed the case, but I don't think it's a big enough problem to start duplicating `toString` definitions for.

PR #810, opened 2016-07-31
Update CONTRIBUTING.md

apinder:


marijnh:
Thanks for noticing that. Merged as ce2c135

PR #812, opened 2016-07-31
Add lookup_paths option to node_resolve plugin

liberlanco:
We use NODE_PATH env variable in our project, it contains several specific "libs" paths where we store all libraries. Inside the code we load them with short line `var lib = require('project/lib');`
Tern failed to find them because node_resolve plugin looks for files only in standard paths.

This patch adds ability to configure project specific paths for node_resolve plugin. For example .tern-project:

``` json
  {
    "libs": [
        "underscore"
    ],
    "plugins": {
        "node_resolve": {
            "lookup_paths": [
                "lib",
                "3rdparty/lib"
            ]
        },
        "node": {}
    }
  }
```

Works fine with tern_for_sublime.

### DIFF
--- a/plugin/node_resolve.js
+++ b/plugin/node_resolve.js
@@ -15,6 +15,7 @@
   function findDeclaredDeps() {}
 
   var resolveToFile
+  var projectPaths = [];
   if (require) (function() {
     var module_ = require("module"), path = require("path"), fs = require("fs")
 
@@ -27,7 +28,9 @@
 
       var parentModule = {
         id: fullParent,
-        paths: module_._nodeModulePaths(parentDir).concat(module_.globalPaths)
+        paths: module_._nodeModulePaths(parentDir)
+                      .concat(projectPaths)
+                      .concat(module_.globalPaths)
       }
       try {
         return module_._resolveFilename(name, parentModule)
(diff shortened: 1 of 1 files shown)


marijnh:
Wouldn't it be an option to simply pass the same environment variable when you run the Tern server? I'm not too keen on adding even more poking into undocumented Node internals here.

liberlanco:
There were 3 variants where to place those paths

1. In environment before I start editor (link or shell wrapper script)
2. In term_for_sublime, so it will read per-project settings and setup env before spawning tern server.
3. In per project setting.

1st one gives most of conflicts between projects. I have either to have several per-project links to open same editor to be sure when I edit one project, libs from another won't be read because of global env var.

2nd one - I actually tried in tern_for_sublime, but stopped because I failed to find out how to make it read per-project settings. Global ones gives same problem as in 1st case. Also downside of this - it will work only in sublime, other editors need coding too.

3rd one - looks like most right place to write this feature, logically. But it's implemented via hack, node doesn't have feature to update lookup paths which exists long time in python or ruby. What to do ...

If it's not acceptable, I'd voted for rewriting whole node_resolver module to spawn separate process with corrent environment and do lookup via it, instead of 1st or 2nd variants.

liberlanco:
I have rewritten patch to not use Node internals

plugin/node_resolve.js
```diff
@@ -15,6 +15,7 @@
   function findDeclaredDeps() {}
 
   var resolveToFile
+  var projectPaths = [];
```
marijnh: Put this in the `Modules` object, not as a file-global variable. (You could have multiple Tern servers active in a process, with different configurations.)

marijnh:
Closing due to inactivity. Feel free to submit a new PR if you do more work on this.

prog666:
I need this feature, did you figure out some other way to change NODE_PATH?

prog666:
I did try to apply this changes, it works only if you pass absolute path to config, like "/Users/username/projects/myproject/src"

prog666:
It does work, but slows down autocomplete terribly

PR #813, opened 2016-08-05
Minor memory enhancement of dropping previous requests

FrancisMurillo:
I like the autocomplete support of Tern with Emacs. When I use it with `company`, I seem to hang from time to time whenever it makes a request to tern, it could be my just my hardware but I'm trying to pinpoint why it is choking up from time to time.

This pull request simply ignores the previous request made from being processed or triggered. This slightly helps when I am typing faster than it completes so it prevents the previous completions from triggering.


emacs/tern.el
```diff
@@ -293,15 +293,17 @@ list of strings, giving the binary name and arguments.")
...
-                            (when type
-                              (setf tern-last-argument-hints (cons opening-paren type))
-                              (tern-show-argument-hints))))
-                        `((type . "type")
-                          (preferFunction . t))
-                        opening-paren
-                        :silent)))))
+        (lexical-let ((generation tern-command-generation))
+          (tern-run-query (lambda (data)
+                            (when (= tern-command-generation (1- generation))
```
marijnh: Isn't `tern-run-query` already doing this test?

FrancisMurillo: My memory fails me on this. 
I remember I was being aggressive in denying previous requests from executing.
Testing out the code again, you might be right.

marijnh: Could you adjust the PR to revert this part?

FrancisMurillo:
Removed redundant aggressive conditional

marijnh:
Thanks. Merged as 391bb28

Issue #814, opened 2016-08-11
Feature request: async/await support


rkistner:
I realize that it is not standardized yet (currently stage 3), but support for async/await would be extremely useful for me. What would it involve to get this implemented?

The functionality I'm looking for would entail:

``` js
// Case 1
async function myfunction() {
    return 'test';
}

var promise = myfunction();
// promise inferred as Promise<string> (currently just string).

// Case 2
async function myfunction2() {
    var promise = Promise.resolve('test');
    var value = await promise;
    // value inferred as string (currently unknown).
}
```

I'm willing to help with the implementation in my spare time. If you feel this should not be in the core library since it's not standardized yet, would it be feasible to implement this in some kind of "plugin", or would I have to fork ternjs? I understand that it is also not part of the acorn parser yet, but there already exists plugins an acorn plugin implementing it.

rkistner:
Alternatively, instead of implementing it in ternjs directly, would it be feasible to "pre-transform" the AST before passing it to tern? I can imagine something like the following:

1. Inside any async function, transform `return <expression>` into `return Promise.resolve(<expression>)`.
2. Transform `await <expression>` into `Promise.resolve(<expression>)._value` (not sure exactly how to access the underlying value type here, but should be able to figure something out).

Of course these transformations are very different in behaviour, but should give the same result for type inference.

rkistner:
I have a basic proof of concept working now with the transforms as described above, so the approach seems to be working. There are still some tricky cases such as arrow functions, but I should be able to figure something out. I'll share my code when I've cleaned it up.

So it does seem like it's feasible to implement this completely outside of Tern, but the implementation would be cleaner if Tern could work with e.g. `AwaitExpression` directly.

rkistner:
Initial proof of concept here: https://github.com/rkistner/tern-asyncawait. Tests are passing, but haven't actually used it yet.

marijnh:
This was landed with #446

rkistner:
https://github.com/ternjs/acorn/pull/446 only implements it in Acorn, not in Tern yet, so this should probably stay open for now?

marijnh:
Oops, I thought this was in the Acorn repo. Reopening!

CandySunPlus:
+1

jsl9208:
Any progress?

peltsi212:
Node 8 has been released 2017-05-30 and it supports async/await
eslint support ES2017
[...]

PR #815, opened 2016-08-12
Fixes for Promise.resolve

rkistner:
Say you have `p = Promise.resolve('a string')`, then `p` is a `Promise<String>`. The issue here is that `Promise.resolve(p)` is inferred as `Promise<Promise<string>>`, where it should be `Promise<string>`.

These two cases specifically were broken (the other added cases are to check for potential issues in my implementation):

```js
var p4 = Promise.resolve(Promise.resolve(10));
p4.then(function(value) {
  value; //: number (was Promise<number>)
});

var arg5 = 1 < 2 ? Promise.resolve(10) : 20;
var p5 = Promise.resolve(arg5);
p5.then(function(value) {
  value; //: number (was Promise<number>|number)
});
```

This PR fixes these test cases, but I may be missing some other edge cases (see the TODOs in the code).

It's a fairly niche use-case, but can make a big difference when working with Promise-heavy code (especially with async/await, which I'm experimenting with now - see #814).

### DIFF
--- a/defs/ecma6.json
+++ b/defs/ecma6.json
@@ -680,7 +680,7 @@
     },
     "reject": "Promise_reject",
     "resolve": {
-      "!type": "fn(value: ?) -> +Promise[:t=!0]",
+      "!type": "fn(value: ?) -> !custom:Promise_resolve",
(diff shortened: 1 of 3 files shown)


marijnh:
Thanks. Merged as 1cba2cd and b0c3f48c, brought up to date in 7effbec20db952

rkistner:
Thanks!

PR #818, opened 2016-08-17
Issue 817 - add Geolocation to browser.json

mrennie:


defs/browser.json
```diff
@@ -3379,5 +3384,110 @@
     "!type": "fn(width?: number, height?: number) -> +HTMLImageElement",
     "!url": "https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image",
     "!doc": "Image Element constructor. Accepts two optional parameters: Image([unsigned long width, unsigned long height]). Returns an HTMLImageElement instance just as document.createElement('img') would."
+  },
+  "Geolocation": {
```
marijnh: These aren't actual global variables, so I'd prefer you put them in the `!define` block, so that you can reference them, bu they won't show up as completions.

marijnh:
Thanks! Merged as 675165175 and followed up with 0bf87473b because the plus sign will create an instance of the prototype of something, and your interface definitions don't have prototypes.

PR #821, opened 2016-08-29
Update some Node.js definitions

TimothyGu:


TimothyGu:
As a question, is there a way to define the keys of an object? Like `fn(filename: string, options?: {option1: number})`

marijnh:
The syntax you show should work (a braced expression is a valid Acorn type). You can, if you want to reuse a type, also add it to the file's `!define` section and refer to it by name.

Why'd you shuffle the order of the properties around (which makes it hard to read the patch)?

Also, what's with all the square brackets in the updated docstrings?

TimothyGu:
> Why'd you shuffle the order of the properties around (which makes it hard to read the patch)?

I'm generating the definitions using a script, and object properties are not ordered in the same way in Node.js documentation as the already existing definitions. I'll find a way to make it easier to read.

> Also, what's with all the square brackets in the updated docstrings?

Hmm I didn't notice that. I'll fix it.

TimothyGu:
> The syntax you show should work (a braced expression is a valid Acorn type).

Seems like the type parser doesn't support it:

```
/home/timothy-gu/tern/tern/lib/def.js:85
      throw new Error("Unrecognized type spec: " + this.spec + " (at " + this.pos + ")");
      ^

Error: Unrecognized type spec: fn(filename: string, options?: {option1: number}) (at 31)
    at Object.TypeParser.error (/home/timothy-gu/tern/tern/lib/def.js:85:13)
    at Object.TypeParser.parseFnType (/home/timothy-gu/tern/tern/lib/def.js:103:33)
    at Object.TypeParser.parseTypeInner (/home/timothy-gu/tern/tern/lib/def.js:160:21)
    at Object.TypeParser.parseTypeMaybeProp (/home/timothy-gu/tern/tern/lib/def.js:146:25)
    at Object.TypeParser.parseType (/home/timothy-gu/tern/tern/lib/def.js:130:23)
    at parseType (/home/timothy-gu/tern/tern/lib/def.js:262:59)
    at passTwo (/home/timothy-gu/tern/tern/lib/def.js:424:9)
    at passTwo (/home/timothy-gu/tern/tern/lib/def.js:442:11)
    at passTwo (/home/timothy-gu/tern/tern/lib/def.js:442:11)
    at doLoadEnvironment (/home/timothy-gu/tern/tern/lib/def.js:484:5)
```

marijnh:
Oh, you're right, `{prop: type}` isn't parsed by the JSON definition parser. I guess I was thinking of the JSDoc parser. So you'll have to define the object type in your `"!define"` section.

TimothyGu:
@marijnh, could you take a look at this when you get the time, please?

marijnh:
Sure. Merged as dcd808f985e7...435b92882d290

PR #823, opened 2016-08-30
Support rest parameters and object literals in defs

TimothyGu:
Rest parameters are relatively straightforward. For object literals, I wasn't entirely certain what the `computed` boolean is for, and how would I use it. Hence the WIP tag.

### DIFF
--- a/lib/def.js
+++ b/lib/def.js
@@ -68,6 +68,15 @@
       return new infer.Arr(types.map(function(tp) { return unwrapType(tp, self, args) }))
     }
   }
+  function computedObject(names, types) {
+    return function(self, args) {
+      var obj = new infer.Obj;
+      names.forEach(function (prop, i) {
+        obj.defProp(prop).addType(unwrapType(types[i], self, args));
+      });
+      return obj;
+    };
+  }
 
   TypeParser.prototype = {
     eat: function(str) {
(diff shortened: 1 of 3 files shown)


lib/def.js
```diff
@@ -173,6 +181,20 @@
...
         }
         return new infer.Arr(types || inner)
+      } else if (this.eat("{")) {
+        var types = [], names = [], computed = false
+        if (!this.eat("}")) {
+          computed = this.parseList(comp, types, names)
+          this.eat("}") || this.error()
+        }
+        if (computed) {
+          this.error();
```
TimothyGu: Specifically, here: I'm not sure how to handle `computed`.

marijnh: You'll have to write a `computedObject` function analogous to `computedArray`, an call that. These are for return types that depend on the function's input type (i.e. you could say `fn(arg: ?) -> {content: !0}` for a function that wraps its argument in an object.

lib/def.js
```diff
@@ -173,6 +181,20 @@
           return this.base
         }
         return new infer.Arr(types || inner)
+      } else if (this.eat("{")) {
+        var types = [], names = [], computed = false
+        if (!this.eat("}")) {
+          computed = this.parseList(comp, types, names)
```
marijnh: This will allow triple dots in property names, though. And it'll disallow quoted property types, etc. Not sure if the logic for parsing object fields and function arguments is really similar enough.

TimothyGu:
Comments are addressed.

marijnh:
Thanks. Merged as 435b928..a36a20c

PR #826, opened 2016-09-06
Resolve .jsx files

lentan1029:
If you wish to ignore .jsx files, you can explicitly define that in a regex in the dontLoad option

### DIFF
--- a/plugin/modules.js
+++ b/plugin/modules.js
@@ -64,7 +64,7 @@
       var known = this.modules[resolved]
       if (known) return known
 
-      if (/\.js$|(?:^\/)[^\.]+$/.test(resolved))
+      if (/\.jsx?$|(?:^\/)[^\.]+$/.test(resolved))
         this.server.addFile(resolved, null, parentFile)
       if (!relative) this.nonRelative[name] = resolved
       return this.modules[resolved] = new infer.AVal
@@ -116,7 +116,7 @@
       var path = parentFile ? resolvePath(dirName(parentFile), word) : baseName(word)
       for (var prop in this.modules) {
         if (prop != parentFile && filter(path, prop, query)) {
-          if (/\.js$/.test(prop)) prop = prop.slice(0, prop.length - 3)
+          if (/\.jsx?$/.test(prop)) prop = prop.slice(0, prop.length - 3)
           var added = prop.slice(path.length)
           tern.addCompletion(query, completions, word + added, this.modules[prop])
         }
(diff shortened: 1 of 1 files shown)


JelteF:
I tried this, but I can't get it to work. What does your  `.tern-project` file look like?

plugin/modules.js
```diff
@@ -116,7 +116,7 @@
       var path = parentFile ? resolvePath(dirName(parentFile), word) : baseName(word)
       for (var prop in this.modules) {
         if (prop != parentFile && filter(path, prop, query)) {
-          if (/\.js$/.test(prop)) prop = prop.slice(0, prop.length - 3)
+          if (/\.jsx?$/.test(prop)) prop = prop.slice(0, prop.length - 3)
```
chessman: There must be prop.length - 4 because length(".jsx") = 4.

JelteF: It can be either 3 or 4, so jou would need two different if statements

chessman: and 'else if' is also appreciated because it won't work with "foo.jsx.js" file name.

marijnh: Preferably, just store the regexp match and use `match[0].length` to get its actual length

markwain:
Hi!
Any updates on this issue? ... I can't use tern because I have a lot of .jsx files that he simply doesn't see.

markwain:
@marijnh Can I help with something ... maybe update PR? I really like tern and would like to use it :)

JelteF:
I just used the webpack plugin as I use that anyway, so I didn't have a need for this anymore.

markwain:
@JelteF can you describe it more?
Because I also use webpack plugin ... but it doesn't help.

So my .tern-project files looks like this:
```
{
  "plugins": {
    "webpack": {
      "configPath": "./client/webpack.config.js"
    },
  },
  "libs": [
    "browser"
  ],
  "ecmaVersion": 6
}
```
webpack.config:
```
resolve: {
    extensions: [
      '*',
      '.js',
      '.jsx'
    ],
    alias: {
      actions: srcPath + '/actions/',
      components: srcPath + '/components/',
      containers: srcPath + '/containers/',
      reducers: srcPath + '/reducers/',
      store: srcPath + '/store/',
      images: srcPath + '/images',
      config: srcPath + '/config',
      configuredLibraries: srcPath + '/configuredLibraries',
      utils: srcPath + '/utils.js',
    },
  },
```

Webpack plugin helps tern resolve webpack's aliases, but tern still doesn't see .jsx files. Though it became to see them after .jsx files got edited and saved.

plugin/modules.js
joeheyming: I found replace to work well here:
e.g.
prop = prop.replace(/\.jsx?/, '')

PR #839, opened 2016-10-16
Update doc_comment plugin to handle obj properties

### LINKED ISSUE #673: JSDoc support for parameters with properties
It should b ecool if ternjs could support JSDoc for parameters with properties http://usejsdoc.org/tags-param.html#parameters-with-properties

dannysu:
Fixes #673.


marijnh:
Thanks! Merged as 4a5fe8f

PR #840, opened 2016-10-17
Fix !url in angular plugin

dannysu:


marijnh:
Thank you. Merged as e262991

PR #843, opened 2016-10-30
Added optimal travis config

amilajack:


marijnh:
We're still intending to run on node 0.10, so I want to keep that as a target.

amilajack:
Fixed

marijnh:
Thanks. Merged as 9d06ec7

Issue #844, opened 2016-11-03
documentation request does not work for ClassDeclarations


mrennie:
Consider the following snippet:

```
/**
 * @description This is a class decl that I figured I should talk about
 * @name MyClass
 * @since 1.0
 */
class MyClass {
	constructor() {}
}
export default MyClass;
export let instance = new MyClass();
```

If you issue a documentation request with the position in any of the 'MyClass' identifiers, no documentation is returned.

Related Orion bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=507000

PR #845, opened 2016-11-03
documentation request does not work for ClassDeclarations #844

mrennie:
We just have to make sure we attach the comments to the declaration node, and avoid running isFnExpr on the ClassDeclaration node


marijnh:
Oh, woops, I pushed a patch before I saw you had opened a PR. I didn't cover one case, so I've merged what's left of your patch as well, as 0c5c1d2

Issue #849, opened 2016-11-19
Add HTMLElement.matches() method.


yaphtes:
Plese, add HTMLElement.matches() method.
https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

marijnh:
Want to submit a pull request? (Add it to defs/browser.json)

PR #853, opened 2016-11-28
Add HTMLElement.matches() method. #849

mrennie:
Adds the matches function to the Element prototype in browser.json


marijnh:
Thanks! Merged as 8e2e0ca

Issue #850, opened 2016-11-19
Please add methods.


yaphtes:
Hello.
Can you add next methods?
`Element.prototype.append();`
`Element.prototype.prepend();`
`Element.prototype.before();`
`Element.prototype.after();`

I don't know `prototype`, maybe is:
`Node.prototype.append();`
`Node.prototype.prepend();`
`Node.prototype.before();`
`Node.prototype.after();`

marijnh:
Could you submit a pull request?

yaphtes:
no i can`t, because i am new member in github.
UPD: plese, add methods, elem.matches() and elem.closest()
links in mdn: https://developer.mozilla.org/en/docs/Web/API/Element/closest and https://developer.mozilla.org/en/docs/Web/API/Element/matches

mrennie:
I can submit a PR for this. Just one question though, how do you represent varargs in an index?

For example, the API doc for append(...) is a list of strings or nodes. So would the index type look like: `fn(nodes: string|Node?)`

yaphtes:
I think your API doc is well suited. 👍

marijnh:
You should be able to write a type like `fn(...nodes: [string | Node])`

PR #855, opened 2016-11-30
Please add methods. Issue #850

mrennie:
Adds the append, prepend, before, after and closest methods to the index.

In the doc for the new methods I added a blurb: "This is an experimental technology" since each of the methods is tagged as experimental in the MDN docs.


marijnh:
Merged as ca40568 (after fixing a syntax error in a type).

PR #857, opened 2016-12-05
Prevent iteration for <top> proto

chemzqm:
I found that It would cost 100% CPU usage when iterate through the infer named `<top>`, since it would have lot's of props.

I run into this issue with a es6 file which have many `export` statements.

screen shot 2016-12-04 at 4 25 41 pm

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -656,11 +656,14 @@
         if (this.proto) this.proto.forAllProps(this);
       }
       this.onNewProp.push(c);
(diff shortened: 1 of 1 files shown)


marijnh:
This will break completion of top-level variables. I don't see how exports have anything to do with the top scope, nor do I understand how you'd get so many top level variables that iterating through the top scope becomes this expensive (it'd have to be, well, millions of them).

chemzqm:
I've find out the issue is with parsing jsx files, the parse would be stumped if there are many jsx blocks.

PR #859, opened 2016-12-12
Documentation of webpack plugin

chemzqm:
I can't use make to generate the manual.html file, looks like my source-highlight not configured correctly, and I don't know how to make it works.

```
source-highlight: missing feature: language inference requires input file
source-highlight: could not find a language definition for application/x-json
asciidoc: WARNING: manual.txt: line 504: filter non-zero exit code: source-highlight -f xhtml -s application/x-json: returned 1
asciidoc: WARNING: manual.txt: line 504: no output from filter: source-highlight -f xhtml -s application/x-json
```

I've noticed there're some break change in enhanced-resolve v3.0.0, I think it's better to wait for webpack use it instead of upgrade too early.

### DIFF
--- a/doc/manual.txt
+++ b/doc/manual.txt
@@ -670,6 +670,37 @@ This plugin (`es_modules`) builds on top of the
 <<plugin_modules,`modules` plugin>> to support ECMAScript 6's `import`
 and `export` based module inclusion.
 
+[[plugin_webpack]]
+==== Webpack plugin ====
+
+This plugin (`"webpack"`) make use of `enhance-resolve` module from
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "acorn": "^3.2.0",
-    "enhanced-resolve": "^2.2.2",
+    "enhanced-resolve": "^2.2.0",
```
marijnh: Is this intentional? If so, why?

chemzqm: It's not necessary, I'll remove that.

marijnh:
Done: c752a24

PR #870, opened 2016-12-24
Allow for webpack function configs

eloytoro:
Webpack allows configs to be functions, this would enable support for that

### DIFF
--- a/plugin/webpack.js
+++ b/plugin/webpack.js
@@ -31,6 +31,8 @@ function getResolver(configPath) {
         config.mainFields = resolveConfig[key]
       } else if (key === 'root') {
         config.modules.unshift(resolveConfig[key])
+      } else if (key === 'modules') {
+        config.modules = config.modules.concat(resolveConfig[key])
       } else {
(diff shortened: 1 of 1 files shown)


chemzqm:
Looks good, didn't notice there's function config.

eloytoro:
Decided to add the ability to just pass the modules you're trying to resolve.
My webpack config does some heavy computing by just requiring it and it was causing lots of lag.
I think its great to have it for simplicity as well

chemzqm:
@eloytoro the `modules` are folder for resolve modules, which is just node_modules most of case, why your webpack need computed for that?

eloytoro:
remember we're loading the entire webpack config, my project is quite big and the config has a lot of dependencies, thus requiring the config means requiring a bunch of stuff and sometimes that means heavy computation.

> which is just node_modules most of case

This is, as far as as I've seen, in practice not accurate, most projects have more than one resolve directories in their config

marijnh:
Sounds reasonable. Merged!

Issue #877, opened 2017-01-17
browser index items for Storage, localStorage and sessionStorage are not correct


mrennie:
The index is completely missing the Storage interface, which localStorage and sessionStorage are implementations of. They are also missing clear(), length, get and removeItem.

I'll open a PR to update the index.

marijnh:
Resolved via #878

PR #878, opened 2017-01-17
browser index items for Storage, localStorage and sessionStorage are not correct

mrennie:
Fixes issue #877 - adds in the missing Storage type and members


marijnh:
Thank you. Merged as 4a88712

Issue #883, opened 2017-01-29
Webpack plugin doesn't resolve in other roots


LubergAlexander:
Minimal repo to reproduce: https://github.com/LubergAlexander/ternjs-webpack-resolver-bug

`npm i && npm run build` builds the bundle successfully using the webpack config in the repo.

`tern --verbose` shows an exception when trying to resolve the module using webpack:

```
Error: Can't resolve 'utils/testUtil' in '/Users/aluberg/Code/tern_webpack_resolve'
    at onResolved (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/Resolver.js:66:16)
    at loggingCallbackWrapper (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/createInnerCallback.js:31:19)
    at afterInnerCallback (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/Resolver.js:138:10)
    at loggingCallbackWrapper (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/createInnerCallback.js:31:19)
    at Resolver.applyPluginsAsyncSeriesBailResult1 (/usr/local/lib/node_modules/tern/node_modules/tapable/lib/Tapable.js:181:46)
    at innerCallback (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/Resolver.js:125:19)
    at loggingCallbackWrapper (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/createInnerCallback.js:31:19)
    at /usr/local/lib/node_modules/tern/node_modules/tapable/lib/Tapable.js:283:15
    at /usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/UnsafeCachePlugin.js:39:4
    at loggingCallbackWrapper (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/createInnerCallback.js:31:19)
Error: Can't resolve 'utils/sampleModule' in '/Users/aluberg/Code/tern_webpack_resolve'
    at onResolved (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/Resolver.js:66:16)
    at loggingCallbackWrapper (/usr/local/lib/node_modules/tern/node_modules/enhanced-resolve/lib/createInnerCallback.js:31:19)
[...]

PR #884, opened 2017-01-29
Fix root array not works with webpack resolve

chemzqm:
Fix of #883, also add support for fallback field of webpack resolve.

http://webpack.github.io/docs/resolving.html

### DIFF
--- a/plugin/webpack.js
+++ b/plugin/webpack.js
@@ -6,6 +6,10 @@ var tern = require("../lib/tern");
 require("./commonjs");
 require("./es_modules");
 
+function isArray(v) {
+  return Object.prototype.toString.call(v) == "[object Array]";
+};
+
 var fs = require('fs');
 var path = require("path");
 var ResolverFactory = require("enhanced-resolve").ResolverFactory;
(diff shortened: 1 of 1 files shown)


plugin/webpack.js
```diff
@@ -6,6 +6,10 @@ var tern = require("../lib/tern");
 require("./commonjs");
 require("./es_modules");
 
+Array.isArray = Array.isArray || function(v) {
```
marijnh: Please don't patch global objects from the library. We can use a local `isArray` variable for this.

LubergAlexander: I would like to get this merged, how can I help ? @chemzqm would it be possible for you to fix the comment by @marijnh ? I can do that, but I will probably have to copy your changes to my own branch and issue another pull request

chemzqm: I've fixed that problem, I think it's ready get merged.

LubergAlexander: @chemzqm thanks! that was fast! @marijnh is it possible to merge it ?

marijnh:
Thanks. Merged as db13aac

PR #886, opened 2017-02-06
update acorn dependency

ramitos:


marijnh:
Enhanced-resolve seems to have started shipping ES6 code, which breaks node versions that Tern still wants to support.

ramitos:
ah, good point :( I tested locally, but I'm running a late version of node.

Just updated the PR to only upgrade acorn

marijnh:
Thanks. Merged

PR #890, opened 2017-02-19
Implement navigate through symbol occurrences in emacs

### LINKED ISSUE #723: Highlight and navigate through symbol occurrences on emacs module
I recently found http://mihai.bazon.net/projects/editing-javascript-with-emacs-js2-mode/js2-highlight-vars-mode, which allows me to easily see all instances of the variable at point and also allows me to navigate between them. This is quite convenient, and it seems like something tern could do.

My current problem with the above implementation is that it interferes with company-mode completions. Suggestions only give me about a second to tab complete before they dissappear (i'm guessing due to the 'polling' mechanism used by that plugin). I can't seem to find where I could contact that developer, so I figured I'd go ahead and make a feature request here first since I'd rather just use one tool.

aiguofer:
Inspired by js2-highlight-vars-mode, navigate through ref
occurrences. Fixes #723 

Function names and key bindings not set in stone, so let me know if you have any feedback!

### DIFF
--- a/emacs/tern.el
+++ b/emacs/tern.el
@@ -412,16 +412,53 @@ list of strings, giving the binary name and arguments.")
     (run-with-timer tern-flash-timeout nil 'delete-overlay overlay)))
 
 (defun tern-do-highlight (data)
-  (cl-loop for ref across (cdr (assq 'refs data)) do
-           (let ((file (cdr (assq 'file ref))))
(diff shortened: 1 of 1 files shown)


ArneBab:
This looks like exactly what I need just now. Is there a reason why it has not been merged?

Issue #891, opened 2017-02-24
TypeError: Cannot read property 'prev' of null


dbellavista:
Hi!
I'm using tern `0.21.0` on a MacOs system, feeding it with the vim plugin deoplete-ternjs.

I got this error during the autocompletion:

```
/usr/local/lib/node_modules/tern/lib/infer.js:988
        var scopeCopy = new Scope(scope.prev, scope.originNode);
                                       ^

TypeError: Cannot read property 'prev' of null
    at /usr/local/lib/node_modules/tern/lib/infer.js:988:40
    at withDisabledComputing (/usr/local/lib/node_modules/tern/lib/infer.js:359:18)
    at fn.computeRet (/usr/local/lib/node_modules/tern/lib/infer.js:984:14)
    at ctor.addType (/usr/local/lib/node_modules/tern/lib/infer.js:382:18)
    at withWorklist (/usr/local/lib/node_modules/tern/lib/infer.js:908:19)
    at exports.AVal.propagate (/usr/local/lib/node_modules/tern/lib/infer.js:94:25)
    at Modules.<anonymous> (/usr/local/lib/node_modules/tern/plugin/commonjs.js:71:17)
    at Modules.signal (/usr/local/lib/node_modules/tern/lib/signal.js:28:49)
    at exports.Server.<anonymous> (/usr/local/lib/node_modules/tern/plugin/modules.js:338:24)
    at exports.Server.signal (/usr/local/lib/node_modules/tern/lib/signal.js:28:49)
```

chemzqm:
It should be problem of deoplete-ternjs, you can use official vim plugin https://github.com/ternjs/tern_for_vim and bind ominicomplete to deoplete.

marijnh:
I really doubt this has anything to do with deoplete-ternjs. It looks like an issue with the inference engine.

Do you have an example of a file that triggers this error?

dbellavista:
@marijnh I noticed the error only after a while. I'll try to reproduce it but I can't make any promise :)

camdenb:
I'm having this error consistently. Running YouCompleteMe with tern on Vim 8.0.

jsinai:
I am seeing this problem in atom, consistently.

lyjstudy:
FIXED: https://github.com/lyjstudy/tern
rmScope will destroy node.scope, so we cache it

marijnh:
Solved via #905

PR #897, opened 2017-04-27
Fix `Cannot read property 'prev' of null`

jfelsinger:
Not sure if this is the correct way to do this, but it fixed the bug that I was seeing locally with auto-completion and all that.

Should be a fix for: #891

### DIFF
--- a/lib/infer.js
+++ b/lib/infer.js
@@ -987,7 +987,7 @@
         var oldOrigin = cx.curOrigin;
         cx.curOrigin = fn.origin;
         var scope = node.scope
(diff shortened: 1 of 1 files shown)


marijnh:
That doesn't really address the underlying issue, though. I'd prefer to see an example of how to reproduce it so that I can properly fix it.

guiguan:
This issue can be reproduced with `atom-ternjs` package. Probably @jfelsinger can provide a concrete example?

lyjstudy:
FIXED: https://github.com/lyjstudy/tern
rmScope will destroy node.scope, so we cache it

marijnh:
Solved via #905

PR #898, opened 2017-04-27
typo in demo

berkin:
i just keep trying to check the jump the definition with ctrl key. small typo is fixed in demos.


marijnh:
Woops. Thanks for spotting that. Merged.

PR #905, opened 2017-05-11
add refScope

lyjstudy:
function rmScope(node) { if (node.scope) node.scope = null } destroy node.scope
we ref scope in setFunctionInstantiated


marijnh:
Thanks for tracking that down. It's not great that the ordering of type inference is so messy that this actually ends up running before the scope is populated, but I guess that's what happens, and this seems a decent workaround. Merged.

PR #908, opened 2017-05-18
Allow for webpack resolve configs to be passed directly

eloytoro:
Right now term supports passing the `configPath` and the `modules` options to the webpack plugin.

As it so happens, webpack 2 natively supports configs to be written using ES6 modules because it transpiles the config using babel if it has the `.babel.js` termination.

There are two ways to solve this as far as I can see.

* Add babel-core as a dependency to tern
* Allow to pass the values directly to the config

The first solution is the best one for users in my opinion, being able to require their babel-esque config would silently fix the issue. If you guys like it we can close this PR and I can open another one that takes care of this.

The second solution is what I'm doing here, allowing to pass `"alias"` to the config should fix the issue for now

### DIFF
--- a/doc/manual.txt
+++ b/doc/manual.txt
@@ -702,6 +702,23 @@ like:
 `configPath` should be a file path relative to `.tern-project`, you
 can omit it if they're in the same folder.
 
+You cal also pass the `"module"` and `"alias"` keys directly instead
+of passing the "configPath"
+
+[source,application/x-json]
+----
+{
+  "libs": [
+  ],
+  "plugins": {
+    "webpack": {
+      "alias": { "lib": "./src/lib" },
+      "modules": ["node_modules", "src"]
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


eloytoro:
Note: allowing to compile the config using babel-register adds a noticeable delay to server startup

tyrocca:
Until this is merged, is there a good workaround?

eloytoro:
@tyrocca the only work around would be to not use babel for the webpack config so it can be required directly as a node module

chemzqm:
Another workaround is to make tern read `.babel.js` and use the babel module from the project to compile the config file before load, so no need for add babel-core as a dependency to tern.

eloytoro:
The tern command isn't necessarily executed in the directory where the
project is, so I'm not sure that requiring local dependencies could be
feasible.

Still waiting on a maintainer for a response

PR #912, opened 2017-05-26
Add def file for react

sharils:
This covers React, ReactDOM and PropTypes. However it doesn't cover full API for React since this PR is implemented based on the assumption that the user is writing JSX. So some methods aren't implemented, e.g. React.createElement

---
This change is [Reviewable](https://reviewable.io/reviews/ternjs/tern/912)


marijnh:
Thanks, merged!

PR #915, opened 2017-06-23
Fix lgtm alerts

xiemaisi:
This fixes a few minor problems identified by [lgtm](https://lgtm.com); for more details see https://lgtm.com/projects/g/ternjs/tern/alerts/.

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -389,7 +389,7 @@
 
     for (var i = 0; i < comments.length; ++i) {
       var comment = comments[i];
-      var decl = /(?:\n|$|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)\s+(.*)/g, m;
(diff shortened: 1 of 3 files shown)


plugin/doc_comment.js
```diff
@@ -333,7 +333,6 @@
           var key = parseType(scope, str, pos + 2);
           if (!key) return null;
           pos = skipSpace(str, key.end);
-          madeUp = madeUp || key.madeUp;
```
marijnh: How do you conclude that this statement is useless?

xiemaisi: It's a local variable, and it is reassigned [a few lines down](https://github.com/ternjs/tern/blob/master/plugin/doc_comment.js#L341) without any intervening reads. Or perhaps that assignment was mean to be of the form `madeUp = madeUp || ...` instead?

marijnh: Oh, right, that makes sense. Merging this.

PR #929, opened 2017-09-05
Ensure webpack.js resolves before modules.js

joeheyming:
I had a webpack configuration where modules.js was
resolving node_modules/foo when I wanted my webpack.config.js to
be resolved instead.

In order to enforce the correct order, I changed the
resolve addition to unshift instead of push to the array.
This ensures webpack resolves before modules.

### DIFF
--- a/plugin/webpack.js
+++ b/plugin/webpack.js
@@ -76,7 +76,7 @@ tern.registerPlugin("webpack", function(server, options) {
   var resolver = getResolver(modules, configPath)
   server.loadPlugin("commonjs")
   server.loadPlugin("es_modules")
-  server.mod.modules.resolvers.push(function (name, parentFile) {
+  server.mod.modules.resolvers.unshift(function (name, parentFile) {
     var resolved = resolveToFile(resolver, name, parentFile)
     return resolved && infer.cx().parent.normalizeFilename(resolved)
   })


othree:
Hi, modules.js will use registered resolver first.
If nothing found on any registered resolver.
It will then use default resolver to try again ([ref][1]).

Maybe you use node_resolve plugin too.
And it resolves `node_modules/foo`.
Can you confirm the order of plugins in the `.tern-project` are as expect?

[1]:https://github.com/ternjs/tern/blob/master/plugin/modules.js#L55-L57

othree:
Hi, modules.js uses registered resolvers first.
If nothing found by any registered resolvers.
It will then use the default resolver to try again ([ref][1]).

Maybe you use node_resolve plugin too.
And it resolves `node_modules/foo`.
Can you confirm the order of plugins in the `.tern-project` is as expected?

[1]:https://github.com/ternjs/tern/blob/master/plugin/modules.js#L55-L57

joeheyming:
@othree,  I originally wanted to allow the plugin config have the order.  I can't find the code block I made, but somehow, I decided to switch to this current implementation because the original implementation made no sense to me.  

I'm not very happy working with anyone on this project.  I basically decided to fork the project and go my own way.  I'm still willing to collaborate, but I have been feeling ignored for too long

othree:
Sorry to let you feel bad.
I am in a rush time recently(home, work, family).

I think to change the order is a breaking change.
Might cause other user have issue working on tern + webpack.
So I need take some time to see how people use webpack (I didnt use it for my job and project.)
And I think the worst case is to add a new configure option if this order make any sense.

joeheyming:
I've been using this change for over a year now and it works great

PR #930, opened 2017-09-05
Improved automatic `tern-command' setting of `tern' executable location.

rswgnu:
This is mainly for the case where npm has not installed the executable in a standard executable directory and the tern.el Emacs package has been installed via the Emacs package manager and needs to locate this executable.

### DIFF
--- a/emacs/tern.el
+++ b/emacs/tern.el
@@ -99,12 +99,17 @@
   (let* ((script-file (or load-file-name
                           (and (boundp 'bytecomp-filename) bytecomp-filename)
                           buffer-file-name))
-         (bin-file (expand-file-name "../bin/tern" (file-name-directory (file-truename script-file)))))
-    (if (file-exists-p bin-file)
+         (bin-file (and script-file (expand-file-name "../bin/tern" (file-name-directory (file-truename script-file))))))
+    (unless (or (and bin-file (file-executable-p bin-file))
+		(not (executable-find "npm")))
+      (setq bin-file (or (executable-find "tern")
+			 (expand-file-name "tern/bin/tern"
+					   (file-name-directory (shell-command-to-string "npm -g ls --parseable tern"))))))
+    (if (and bin-file (file-executable-p bin-file))
         (if (eq system-type 'windows-nt) (list "node" bin-file) (list bin-file))
       (list "tern")))
-  "The command to be run to start the Tern server. Should be a
-list of strings, giving the binary name and arguments.")
+  "The command to be run to start the Tern server.
+Should be a list of strings, giving the binary name and arguments.")
 
 (defun tern-start-server (c)
   (let* ((default-directory (tern-project-dir))


emacs/tern.el
```diff
@@ -99,12 +99,17 @@
...
                           (and (boundp 'bytecomp-filename) bytecomp-filename)
                           buffer-file-name))
-         (bin-file (expand-file-name "../bin/tern" (file-name-directory (file-truename script-file)))))
-    (if (file-exists-p bin-file)
+         (bin-file (and script-file (expand-file-name "../bin/tern" (file-name-directory (file-truename script-file))))))
+    (unless (or (and bin-file (file-executable-p bin-file))
+		(not (executable-find "npm")))
+      (setq bin-file (or (executable-find "tern")
+			 (expand-file-name "tern/bin/tern"
+					   (file-name-directory (shell-command-to-string "npm -g ls --parseable tern"))))))
```
marijnh: This just returns `/home/marijn/prog/node/lib` on my system (and tern isn't even installed there). What is it intended to do?

rswgnu:
When the executable 'tern' is not found in its normal place, it is first
looked for within PATH as before and then if not found is searched for
within the npm module directories.  The return value of those lines is not
relevant but what bin-file is set to is.  See what that is on your system
and then look at the lines following the change which determine if the full
path is an existing file or not and then either use that or just use "tern"
with no path.

marijnh:
> where npm has not installed the executable in a standard executable directory

The fallback is already to just run `tern` though, and it seems `start-process` runs it in a way that uses the path, so what exactly does this add (except for the npm-based magic, which seems like it'll easily go wrong if you didn't actually use a global npm install for tern)?

rswgnu:
On Fri, Sep 8, 2017 at 4:02 AM, Marijn Haverbeke <notifications@github.com>
wrote:

> where npm has not installed the executable in a standard executable
> directory
>
> The fallback is already to just run tern though, and it seems
> start-process runs it in a way that uses the path, so what exactly does
> this add (except for the npm-based magic, which seems like it'll easily go
> wrong if you didn't actually use a global npm install for tern)?
>
​The purpose is to find the executable when it has not been installed
within a directory in PATH.
As you note, if you install tern globally, it solves this problem.  If
there are issues with local installs you want to avoid, then just add
conditionals for that.

This is solving a real problem I had when first trying to use tern.  Let's
try to help others avoid the same problem and just handle as many scenarios
as reasonably possible.

Bob

androclus:
I installed both node and tern globally (npm install -g tern) and yet the emacs tern backend still cannot find my tern-server automatically.  I have to set it manually: 

`(setq tern-command '("/home/jas/.nvm/versions/node/v10.13.0/bin/tern"))`

If I do not do this, tern does not "run out of the box" for precisely this use-case (which I would think the most common) that Bob describes: 

- node is not installed in a standard directory. (Many people use nvm, as do I)
- tern.el installed via emacs package manager

Mari, I agree with Bob that this was definitely a stumbling block for me as a new user. It wasted at least a couple hours of my time (I've been working to get tern/emacs going for the 2nd day now.)  And IMHO whenever software can work "out of the box" in at least the most common scenarios, we'll gain more happy users, instead of people who use get frustrated and leave.

Maybe Bob's code above doesn't work perfectly yet, but I think it is the right idea / direction.

ishanch1007 (approved):

PR #933, opened 2017-09-18
Remove needless filename check

AdamNiederer:
When operating on a buffer without an associated file with this check disabled, tern will throw a single error then operate normally. Seems like an improvement over throwing an error each time.


marijnh:
Makes sense -- I think this was added at a point when `tern-project-dir` wouldn't work on file-less buffers.

PR #934, opened 2017-09-19
[angular plugin] Add angular.extend effects

Yabes:
This PR allow completion of properties added with `angular.extend`.


marijnh:
Looks good, merged.

PR #972, opened 2018-07-09
Basic support of async function and await expression.

othree:
This PR includes basic support of async and await.

* async support in `IsCallee`
* await support in `inferExprVisitor`

### DIFF
--- a/bin/tern
+++ b/bin/tern
@@ -50,7 +50,7 @@ var defaultConfig = {
   loadEagerly: false,
   plugins: {doc_comment: true},
   ecmaScript: true,
-  ecmaVersion: 6,
+  ecmaVersion: 8,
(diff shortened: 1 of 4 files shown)


marijnh:
This mostly looks good. But by bumping the `ecmaVersion` option, you are opting into all of the new syntax tree node types that come with ES7 and ES8. There isn't _very_ much new syntax in those versions, I guess, but off the top of my head it means you'll at least have to also implement the `**` exponentiation operator.

othree:
I did some test and `**` work just like `*`. Always return number type.
I will send another PR to include test case for `**`.

Most of other ES7, 8 new features are new methods.
I think its ok to upgrade the default value :)

marijnh:
Sounds good, then!

PR #973, opened 2018-07-13
Add arithmetic operation test

othree:
Add arithmetic operation tests, especially for the new `**` operator.


othree:
Also add test env: node 8, 9.

PR #978, opened 2018-08-08
Async Iteration support

othree:
This PR adds async iteration support to Tern

Changes include:

* Upgrade acorn to 5.4
* Update ecmaVersion to 9
* Add async generator/iterator def
* Handle await for of syntax


marijnh:
Looks good to me!

PR #981, opened 2018-09-11
Fix #871 , findExprAt use new Visitor

### LINKED ISSUE #871: Can't jump to definition for functions exported with shorthand object properties.
Using Tern version 0.20.0 on Emacs 25.1. Node version: 6.9.1

Jumping to definition with `tern-find-definition` when cursor in on a shorthand property doesn't work. Using the regular syntax works. Example:

```js
function create() {
  // Implementation
}

module.exports = {
  create
};
```

The above doesn't work.

```js
function create() {
  // Implementation
}

module.exports = {
  create: create
};
```

The above _works_ correctly when the cursor is positioned on the value word (not the property name).

othree:
findExprAt use fullVisitor, not skip identifier inside object.
findDef/Doc will look around(object) if no result by expr from findExprAt.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -1846,9 +1846,19 @@ scope at that point. Returns <code>null</code> if unsuccessful.
 </dt>
 <dd>
 <p>
-Similar to <a href="#infer.findExpressionAround"><code>findExpressionAround</code></a>,
+Similar to <a href="#infer.findExpressionAt"><code>findExpressionAt</code></a>,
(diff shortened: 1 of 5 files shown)


marijnh:
Property names in member expressions are not expressions, and callers of `findExprAt` will expect to get an expression, so I'm not sure this is a good idea.

othree:
Yes, you are right.
I should create another visitor for `findExprAt` and only walk into `value` of property node.

PR #987, opened 2018-09-14
Upgrade to Acorn 6

marijnh:
This upgrades the library to Acorn version 6, which has some breaking changes.


davidosomething:
it's caret version, but 6.0.1 is out with a fix

PR #1008, opened 2019-03-18
Clarify exp() documentation

byronigoe:
Existing documentation shows "Ex" which implies E*x (E times x). Should be "E^x", which is standard notation for E raised to the x power.


othree:
Thank you

PR #1010, opened 2019-03-19
fix completion for object literal keys as string

### LINKED ISSUE #1009: completions query does not handle strings as object literal keys
example:
```typescript
var obj = {
  Test: "Hello"
};

obj = {
  T // ctrl+space proposes 'Test'
};

obj = {
  "T" // ctrl+space does not propose "Test" (cursor inside the string -> "T|" )
};
```

expected behaviour:
completions query should propose "Test", whether it is a string or not.

nkappler:
fix #1009

### DIFF
--- a/lib/tern.js
+++ b/lib/tern.js
@@ -761,7 +761,7 @@
         var objProp = pointInProp(exprNode, wordEnd);
         if (objProp) {
           objLit = exprAt;
-          prop = isKey = objProp.key.name;
+          prop = isKey = objProp.key.name || objProp.key.value;
         } else if (!word && !/:\s*$/.test(file.text.slice(0, wordStart))) {
           objLit = exprAt;
           prop = isKey = true;


nkappler:
Hi there 😊

Is there any specific reason why this PR receives no attention?
I think the description of the issue ( #1009 ) should be clear and it's a very simple fix.

You can copy and paste the example to the demo page http://ternjs.net/doc/demo/index.html#simple and try it out 😊

othree:
Sorry, I can't reply very quick.
The reason is I still think should completion work in a string literal.
I think current tern don't have any completion suggestion inside a string literal (I might be wrong).

marijnh:
The `complete_strings` plugin completes strings to previously seen strings, and the module plugins will complete known module names, but plain Tern indeed doesn't do anything inside of string literals.

nkappler:
The `complete_strings` plugin is not a solution for this problem in my opinion. It gathers previously seen string literals and completes those, but I am talking about completing the (previously seen) keys of an object literal, which might be written as string (like in JSONs) or not. In the latter case the completion works, in the first one it doesn't.

Here's the example again:
```javascript
var obj = {
  Test: "Hello"
};

obj = {
  T // ctrl+space proposes 'Test'
};

obj = {
  "T" // ctrl+space does not propose "Test" (cursor inside the string -> "T|" )
};
```

nkappler:
Additionally: my fix won't change the behaviour for string literals in general, only in the case mentioned where they are used in an object expression.

PR #1011, opened 2019-03-23
[bugfix] jsdoc - single word tags

itaishmueli:
fixed a bug in the doc_comment plugin, in which tags of a single word would cause the pattern to match two tags at once.
example:
```javascript
/**
 * a test
 * @constructor // single word
 * @param {Object} param1 // would be interpreted as the type of constructor
 * @param {number} param2
 */
function test(param1, param2) {
    return `${param1} - ${param2}`;
}
```
function signature: `fn(param1: ?, param2: number) -> string`
notice the type of param1 is missing because its tag is mis-matched.

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -390,12 +390,14 @@
 
     for (var i = 0; i < comments.length; ++i) {
       var comment = comments[i];
-      var decl = /(?:\n|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)\s+(.*)/g, m;
+      var decl = /(?:\n|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)(?:\s*?\n|\s+(.*))/g, m;
       while (m = decl.exec(comment)) {
         if (m[1] == "class" || m[1] == "constructor") {
           self = foundOne = true;
           continue;
         }
+
+        if (m[2] === undefined) continue; // to avoid tags that require a type argument.
 
         if (m[1] == "this" && (parsed = parseType(scope, m[2], 0))) {
           self = parsed;


othree:
I got this, and looks like its correct:

```
fn(param1: {}, param2: number) -> string
```

Can you make sure you upgrade to latest version?

itaishmueli:
i think it's because i added the extra comments. maybe try this:
```javascript
/**
 * a test
 * @constructor
 * @param {Object} param1
 * @param {number} param2
 */
function test(param1, param2) {
    return `${param1} - ${param2}`;
}
```
but anyway those are my versions of my node and tern installations:
```bash
[***@***** ~]$ jq '.version' /usr/lib/node_modules/tern/package.json 
"0.23.0"
[***@***** ~]$ node --version
v11.11.0
```
this is my .tern-config:
```json
{
    "plugins": {
        "node": {},
        "doc_comment": {
            "strong": true
        }
    }
}
```
from my experience it also happens without `"strong": true`.

othree:
I can reproduce now.
Will do some test to make sure it doesn't break other behavior :)

plugin/doc_comment.js
```diff
@@ -390,13 +390,15 @@
...
       var comment = comments[i];
-      var decl = /(?:\n|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)\s+(.*)/g, m;
+      var decl = /(?:\n|\*)\s*@(type|param|arg(?:ument)?|returns?|this|class|constructor)(?:\s*?\n|\s+(.*))/g, m;
       while (m = decl.exec(comment)) {
         if (m[1] == "class" || m[1] == "constructor") {
           self = foundOne = true;
           continue;
         }
 
+        if (m[2] === undefined) continue; // to avoid tags that require a type argument.
```
othree: Is there any sample for this case?

itaishmueli: this makes sure that a tag without an argument (`@constructor` for example) won't be examined as tags that require one (`@param` for example).

othree: `constructor` already handled at line 395.
So this is just a second guard for any other tags(maybe new tag in the future) without argument?

itaishmueli: It is needed because other tags expect `m[2]` to be a string and it might not be the case. For example, in the param case below it is sliced.
this line continues to the next iteration to avoid an exception.

PR #1014, opened 2019-04-16
 'window.open()' trigger doesn't show DOM window object properties

NehaModi:
It was because browser.json is not having that entry .I have added that entry into json


othree:
Thank you.

Could you change the URL of `!url` to https://developer.mozilla.org/en-US/docs/Web/API/Window/open

othree:
Thank you :)

PR #1015, opened 2019-04-25
When we try to define style for Nodes the hint is not showing up for style

NehaModi:
Changes re done to show the hints for style when defines for Node in the same way as we define for elements

### DIFF
--- a/defs/browser.json
+++ b/defs/browser.json
@@ -194,7 +194,6 @@
         "!url": "https://developer.mozilla.org/en/docs/DOM/Node.textContent",
         "!doc": "Gets or sets the text content of a node and its descendants."
       },
-      "style":  "Element.prototype.style",
       "baseURI": {
         "!type": "string",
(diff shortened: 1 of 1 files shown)


othree:
I know your issue.
But style belongs to `Element` not `Node`.

I prefer replace all returned `+Node` to `+Element` instead of add style to Node, which only adds `style`.

NehaModi:
I have replaced +Node to +element. We have to address this Bug for parentNode ,so I have replaced there.

NehaModi:
ping @othree

othree:
Hi, Sorry I do not reply at the time you send the following commit.

Could update the commit message to meaningful description.

Maybe something like:

> parentNode returns +Element instead of +Node for better usability.

NehaModi:
Hi,
I have changed the commit message as per requirement.Please look into it

NehaModi:
Ping @othree

othree:
Thank you !

PR #1017, opened 2019-05-02
updating Date() object methods

nalkrish:
I have updated Date object methods info according to current standards.

Reference : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date


othree:
@nalkrish Thank you

PR #1023, opened 2019-06-24
Fix TypeError in doc_comment.js

last-partizan:
When editing vue files ternjs is crashing with following error:
```
        var proto = node.objType.getProp("prototype").getObjType();
                                 ^
TypeError: Cannot read property 'getProp' of undefined
```

This change fixed it.

### DIFF
--- a/plugin/doc_comment.js
+++ b/plugin/doc_comment.js
@@ -94,6 +94,7 @@
         }
       },
       Class: function(node, scope) {
+        if (!node.objType) return;
         var proto = node.objType.getProp("prototype").getObjType();
         if (!proto) return;
         for (var i = 0; i < node.body.body.length; i++) {


othree:
Hi, thanks for your PR.
Could you provide an sample code snippets will trigger this issue.
Then I can add it to test cases. (or you can add test case directly to the PR)

last-partizan:
Here is minimal file triggering error:
```
<template>
  <span :is="tag" :class="cls">
    <span :class="cls + '__value'">{{ valueFormatted }}</span>
    <span :class="cls + '__label'">{{ currency.sign + suffix }}</span>
  </span>
</template>
<script>
doc
</script>
```
(cursor at `doc|`)

here is ternjs request:
```
Request: {                                                                                                              
  "query": {
    "type": "completions",
    "types": true,
    "depths": true,
    "docs": true,
    "filter": true,
    "caseInsensitive": false,
    "guess": true,
    "sort": false,
    "expandWordForward": true,
    "omitObjectPrototype": false,
    "includeKeywords": false,
    "inLiteral": true,
    "file": "#0",
    "end": {
      "line": 7,
      "ch": 0
    },
    "lineCharPositions": true
  },
  "files": [
    {
      "type": "full",
      "name": "src/js/widgets/money-widget.vue",
      "text": "<template>\n  <span :is=\"tag\" :class=\"cls\">\n    <span :class=\"cls + '__value'\">{{ valueFormatted }
}</span>\n    <span :class=\"cls + '__label'\">{{ currency.sign + suffix }}</span>\n  </span>\n</template>\n<script>\ndo
\n</script>"
    }
  ]
}
```

othree:
This is because TernJS parse the entire Vue code.
I don't think the fix should be here.
It should modify editor plugins to handle Vue files, and only send the <script> part to the Tern server.

last-partizan:
I started to dig into sources of deoplete-ternjs, and it's possible to do it, but it will require lot more work than adding one line in tern.
And there is a lot editor plugins, this patch can fix them all. (Looks like atom-ternjs has similar error)

Is there any other downsides to this pull-request?

othree:
I think there are two issues:

* Crashing issue
* Editor plugin sends non JS codes to Tern server

This PR fixes the first one.
And my suggestion is for the second.

othree:
Will create a minor release for this fix.

last-partizan:
Thanks.

PR #1024, opened 2019-07-03
Remove note about looking for a maintainer from website

marijnh:
The website still mentions the project is looking for a maintainer. Do you want to remove that?


othree:
Yes, we could remove it.

I am also thinking about we could create a new repo for website.
Host it using github pages, which provides HTTPS service.
And easier for me to update it(don't need to bother you).

What do you think?

marijnh:
Yes, that'd work for me. If you want me to point the domain at some github ip, let me know.

othree:
I will let you know when everything else is ready :)

PR #1037, opened 2020-07-08
Fix typo in property name

bnoordhuis:
Fix casing of the ForInStatement property. Bug appears to have been
introduced in commit 3f4794d ("Clear scopes on AST before analysis")
from December 2016.


othree:
Thanks you

PR #1038, opened 2020-08-14
Adds ecmascr. supp. for TextDecoder, TextEncoder

KartoffelCheetah:
Support is available for both objects in **Nodejs** and **Browsers**.

Browsers make `TextEncoder` and `TextDecoder` available in the global
namespace.

In older versions of **Nodejs (<11)** `TextEncoder` and `TextDecoder` is only
available behind util module. But recent versions of Nodejs moved the
declaration to the global namespace too.

Spec, browser and Nodejs docs linked below.

[1] https://encoding.spec.whatwg.org/#textdecoder
[2] https://encoding.spec.whatwg.org/#textencoder
[3] https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder
[4] https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder
[5] https://nodejs.org/api/globals.html#globals_textdecoder
[6] https://nodejs.org/api/globals.html#globals_textencoder

### DIFF
--- a/defs/ecmascript.json
+++ b/defs/ecmascript.json
@@ -2010,5 +2010,47 @@
         "!url": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet/has"
       }
     }
+  },
+  "TextDecoder": {
+    "!type": "fn(encoding?: string, options?: object)",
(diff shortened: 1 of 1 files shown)


KartoffelCheetah:
It seems that the libraries that are `npm installed` in travis are not compatible with the node version travis is running. Is it a long standing issue?

KartoffelCheetah:
Thank you for the merge othree :) !

